PR #253, opened 2026-05-28
Broadcast permission prompts on permissions:prompt channel

koxx12-dev:
Adds a simple event bus broadcast on "permissions:prompt" channel  for when user is being prompted for permissions.
Very useful for notifications and things of similar nature.

### DIFF
--- a/packages/pi-permission-system/src/index.ts
+++ b/packages/pi-permission-system/src/index.ts
@@ -43,6 +43,7 @@ export default function piPermissionSystemExtension(pi: ExtensionAPI): void {
     subagentSessionsDir: runtime.subagentSessionsDir,
     forwardingDir: runtime.forwardingDir,
     registry: subagentRegistry,
+    events: pi.events,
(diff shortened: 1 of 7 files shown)


gotgenes:
Thank you for the contribution, @koxx12-dev — this feature shipped in `pi-permission-system` v10.0.0 via PR #312.

Your original implementation (commit 8540f3b) was the seed. moekyo picked it up in PR #292 and hardened the payload shape, and we carried it the rest of the way: the channel name settled on `permissions:ui_prompt` (to make clear it fires only when the active user-facing UI is about to be shown, not on every permission request), the payload was trimmed to a lean seven-field contract with a typed `source` discriminant and a `forwarding` context for subagent prompts, and all three permission broadcasts are now best-effort.

The motivating use case you described — notifications that fire exactly when the user needs to return and respond — is exactly what `permissions:ui_prompt` is designed for. The consumer pattern is in `docs/cross-extension-api.md` if it's useful.

PR #292, opened 2026-05-31
feat(pi-permission-system): add UI prompt broadcast contract

moekyo:
## Summary

This builds on the direction from #253 by formalizing a stricter permission UI prompt broadcast contract: `permissions:ui_prompt`.

## Motivation

I am building a notification extension that listens for permission lifecycle events. For permission prompts, `pi-permission-system` should remain the source of truth: notification extensions should not need to patch internals or predict when a prompt will appear from `tool_call` events or permission config.

The important boundary for notification extensions is not "a permission request entered a waiting state". It is "the permission system is about to invoke the active user-facing permission UI, so the user needs to return and respond".

This PR adds a stable event for that narrower boundary so external extensions can notify on the same active permission prompts that `pi-permission-system` actually shows, without duplicating permission logic.

## Relationship to #253

This PR builds directly on #253. The first commit carries the same core idea from that PR: broadcasting before a permission prompt.

The additional commit tightens that contract for notification use cases:

- the public event is now `permissions:ui_prompt`, not a generic request/waiting event
- direct prompts emit only from the active UI prompt path, immediately before `requestPermissionDecisionFromUi`
- non-UI child sessions do not emit when they create forwarded permission requests
- the parent UI session emits before showing the forwarded permission dialog
- the payload is versioned and includes `source`, `surface`, `value`, and related prompt context

If #253 lands first, I am happy to rebase this branch on top of it and keep only the stricter UI-prompt contract and hardening changes. If this lands first, it should cover the same use case as #253 plus the extra contract, forwarded-prompt coverage, and tests.

## Changes

- broadcast `permissions:ui_prompt` before active user-facing permission UI prompts
- keep the event out of auto-resolved paths such as `policy_allow`, `policy_deny`, `session_approved`, `infrastructure_auto_allowed`, and `auto_approved`
- pass direct prompter prompt metadata down into `confirmPermission`, then emit only from the `ctx.hasUI` path
- cover RPC prompts and file-based forwarded permission prompts
- make UI prompt broadcasts best-effort so listener failures cannot block permission handling
- export the UI prompt event contract from the package root
- document the event boundary in the README and cross-extension API reference
- add regression coverage for UI prompt broadcasts, auto-approved paths, and non-UI paths that must not emit

## Third-party listener guidance

Notification extensions should listen to `permissions:ui_prompt` when they need to alert the user:

```ts
eventBus?.on?.("permissions:ui_prompt", handler);
```

`permissions:decision` remains useful for debug logs, status tracking, and audit trails, but it should not be treated as a "notify the user now" event.

## Testing

- `env CI=true pnpm --filter @gotgenes/pi-permission-system test -- test/permission-prompter.test.ts test/permission-forwarding.test.ts test/permission-event-rpc.test.ts test/permission-events.test.ts`
  - 73 files / 1667 tests passed
- `env CI=true pnpm --filter @gotgenes/pi-permission-system check`
  - `tsc --noEmit` passed
- `git diff --check origin/main..HEAD`
  - passed

### DIFF
--- a/packages/pi-permission-system/CHANGELOG.md
+++ b/packages/pi-permission-system/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Features
+
+* broadcast `permissions:ui_prompt` before active user-facing permission UI prompts
+
+### Documentation
+
+* document UI prompt broadcasts in the README and cross-extension API reference
+
 ## [9.2.0](https://github.com/gotgenes/pi-packages/compare/pi-permission-system-v9.1.0...pi-permission-system-v9.2.0) (2026-06-02)
 
 
--- a/packages/pi-permission-system/README.md
+++ b/packages/pi-permission-system/README.md
@@ -20,6 +20,7 @@ Permission enforcement extension for the [Pi](https://pi.mariozechner.at/) codin
 - **Protects sensitive file patterns** — cross-cutting `path` rules deny `.env`, `~/.ssh/*`, etc. across all tools and bash at once
 - **Guards external paths** — prompts before file tools or bash commands reach outside `cwd`
 - **Forwards prompts from subagents** — `ask` policies work even in non-UI execution contexts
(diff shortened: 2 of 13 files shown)


gotgenes:
Thank you, @moekyo — your hardening work in PR #292 is what shaped this into a real contract worth shipping. The `ctx.hasUI` emit boundary, the best-effort try/catch, and the overall direction were all correct; we just carried it the rest of the way. Your authorship is preserved in the commit history on `main`.

Issue #332, opened 2026-06-03
`toolInputPreviewMaxLength` (and `toolTextSummaryMaxLength`) in `config.json` are silently ignored — preview is always truncated at the hardcoded default


kuba-4chain:
### Package

pi-permission-system

### Package version

v10.0.0

### What happened?

Setting `toolInputPreviewMaxLength` (or `toolTextSummaryMaxLength`) in the extension's `config.json` has no effect. The tool-input preview shown in permission prompts is always truncated at the hardcoded default of **200 characters**, regardless of the configured value.

The root cause is a gap between the config loading pipeline and `PermissionSystemExtensionConfig`. The pipeline uses `UnifiedPermissionConfig` (`src/config-loader.ts`) as its intermediate type, which does **not** declare `toolInputPreviewMaxLength` or `toolTextSummaryMaxLength`. As a result:

- `normalizeUnifiedConfig()` never reads these fields from the raw JSON — they are silently dropped.
- `mergeUnifiedConfigs()` iterates only over `["debugLog", "permissionReviewLog", "yoloMode"]` when merging scalar fields, so the values cannot survive the merge even if they were somehow present.
- By the time `normalizePermissionSystemConfig()` is called in `refreshExtensionConfig()`, the merged config object has already lost the fields. `record.toolInputPreviewMaxLength` is `undefined`, `normalizeOptionalPositiveInt(undefined)` returns `undefined`, and `result.toolInputPreviewMaxLength` is never set.
- `resolveToolPreviewLimits()` in `tool-preview-formatter.ts` falls back to the hardcoded constant `TOOL_INPUT_PREVIEW_MAX_LENGTH = 200`.

A secondary bug exists in `saveExtensionConfig()` (`src/runtime.ts`): when the config modal triggers a save, it only writes back `debugLog`, `permissionReviewLog`, and `yoloMode`, so any manually configured `toolInputPreviewMaxLength` / `toolTextSummaryMaxLength` values would be **deleted from `config.json`** silently.

### Expected behavior

When `toolInputPreviewMaxLength: 1000` is set in `config.json`, the inline JSON preview in permission prompts should be truncated at 1000 characters, not 200. The configured value should flow through the load/merge pipeline and reach `ToolPreviewFormatter`.

### Steps to reproduce

1. Set `"toolInputPreviewMaxLength": 1000` in `~/.pi/agent/extensions/pi-permission-system/config.json`.
2. Start an agent session.
[...]

PR #353, opened 2026-06-08
fix(pi-permission-system): preserve runtime knobs through config merge

### LINKED ISSUE #347: piInfrastructureReadPaths in config.json is silently ignored by config-loader merge pipeline
### Package

pi-permission-system

### Package version

10.3.1

### What happened?

I’m trying to allow the agent to read Pi’s installed files without getting an `external_directory` prompt.

   My config includes:

   ```json
   {
     "debugLog": false,
     "permissionReviewLog": true,
     "yoloMode": false,
     "piInfrastructureReadPaths": [
       "~/.local/share/mise/installs/pi"
     ],
     "permission": {
       "external_directory": {
         "*": "ask"
       }
     }
   }
 ```

 But when the agent tries to read a file under that directory, it still goes through the
 external_directory gate:

 ```text
   /read ~/.local/share/mise/installs/pi/0.78.1/README.md
 ```

 Result:

 ```text
   [pi-permission-system] User denied external directory access for tool 'read' path
 '/home/spi-ca/.local/share/mise/installs/pi/0.78.1/README.md'
 ```

 I expected this to be auto-allowed because the path is under `piInfrastructureReadPaths`.

### Expected behavior

 Paths under piInfrastructureReadPaths should bypass the external_directory gate for read-only
 tools like read, grep, find, and ls.

### Steps to reproduce

 1. Add a piInfrastructureReadPaths entry to config.json:

    ```json
      "piInfrastructureReadPaths": [
        "~/.local/share/mise/installs/pi"
      ]
    ```
 2. Keep external_directory set to ask:

    ```json
      "permission": {
        "external_directory": {
          "*": "ask"
        }
      }
    ```
 3. Restart Pi / reload the extension.
 4. Try reading a file under that path:

    ```text
      /read ~/.local/share/mise/installs/pi/0.78.1/README.md
    ```
 5. The external directory prompt still appears.

### Additional context

 I looked through the current code and this seems like it might be related to the config
 loading/merge path rather than the path matching itself.

 normalizePermissionSystemConfig() does handle piInfrastructureReadPaths, but the config seems to
 go through loadAndMergeConfigs() first. The intermediate UnifiedPermissionConfig currently appears
 to preserve only:

 - debugLog
 - permissionReviewLog
 - yoloMode
 - permission

 So by the time normalizePermissionSystemConfig(mergeResult.merged) runs, piInfrastructureReadPaths
  may already have been dropped.

 This looks similar in shape to #332, where another runtime config field was silently ignored by
 the config pipeline.

 Possible places to check:

 - src/config-loader.ts
     - UnifiedPermissionConfig
     - normalizeUnifiedConfig()
     - mergeUnifiedConfigs()
 - src/config-store.ts
     - ConfigStore.refresh()
     - ConfigStore.save()

 A possible fix might be to include piInfrastructureReadPaths in the unified config type, parse it
 in normalizeUnifiedConfig(), and preserve it during merge/save.

spi-ca:
## Summary

- Preserve `piInfrastructureReadPaths` through the unified config load/merge pipeline so configured Pi infrastructure read paths are not silently dropped.
- Preserve optional tool preview runtime knobs during config load, merge, refresh, and save.
- Add regression coverage for runtime knob normalization, merge precedence, refresh behavior, and saving without dropping existing permission config.

## Tests

- `pnpm --filter @gotgenes/pi-permission-system run lint`
- `pnpm --filter @gotgenes/pi-permission-system test`

Closes #347

### DIFF
--- a/packages/pi-permission-system/src/config-loader.ts
+++ b/packages/pi-permission-system/src/config-loader.ts
@@ -21,10 +21,34 @@ export interface UnifiedPermissionConfig {
   debugLog?: boolean;
   permissionReviewLog?: boolean;
   yoloMode?: boolean;
+  piInfrastructureReadPaths?: string[];
+  toolInputPreviewMaxLength?: number;
+  toolTextSummaryMaxLength?: number;
 
   // Flat permission policy
   permission?: FlatPermissionConfig;
 }
+
+const BOOLEAN_RUNTIME_KNOB_KEYS = [
+  "debugLog",
+  "permissionReviewLog",
+  "yoloMode",
+] as const;
+
+const STRING_ARRAY_RUNTIME_KNOB_KEYS = ["piInfrastructureReadPaths"] as const;
+
+const POSITIVE_INT_RUNTIME_KNOB_KEYS = [
+  "toolInputPreviewMaxLength",
+  "toolTextSummaryMaxLength",
+] as const;
+
+const RUNTIME_KNOB_KEYS = [
+  ...BOOLEAN_RUNTIME_KNOB_KEYS,
+  ...STRING_ARRAY_RUNTIME_KNOB_KEYS,
+  ...POSITIVE_INT_RUNTIME_KNOB_KEYS,
+] as const;
+
+type RuntimeKnobKey = (typeof RUNTIME_KNOB_KEYS)[number];
 
 export interface UnifiedConfigLoadResult {
   config: UnifiedPermissionConfig;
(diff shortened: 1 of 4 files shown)


gotgenes:
Thanks for digging into this, @spi-ca — and for the precise diagnosis in both the issue and this PR. You pinpointed the exact gap: `piInfrastructureReadPaths` was declared on `PermissionSystemExtensionConfig` and read at runtime, but `UnifiedPermissionConfig` never parsed or merged it, so it was dropped before `normalizePermissionSystemConfig` ran.

I've shipped a fix in #347, released in `@gotgenes/pi-permission-system` v10.7.0 (51bc145c). It takes the same core approach your PR does — add the field to `UnifiedPermissionConfig`, parse it in `normalizeUnifiedConfig()`, and carry it through `mergeUnifiedConfigs()` with override-wins semantics — with a few differences:

- The `toolInputPreviewMaxLength` / `toolTextSummaryMaxLength` knobs were fixed separately in #332, which has since landed on `main`. That's the main reason this branch now shows conflicts.
- The string-array validator lives in `src/common.ts` (next to `normalizeOptionalPositiveInt`) and is reused to dedupe the existing inline guard in `extension-config.ts`.
- The save path relies on the existing `...existing.config` spread, so once the loader carries the field, the config-modal write-back preserves it without further changes.

One nuance worth flagging from your save-path refactor: writing the knobs from `normalized` rather than `existing.config` means a project-layer `piInfrastructureReadPaths` could be written into the global file on save. I kept the on-disk global value to avoid that, but it's a genuinely under-specified corner — if you think the merged-value behavior is actually what we want, I'm glad to reconsider.

Since the fix is released and the issue is closed, I'm going to close this PR. Thank you for the report and the thorough work on it — much appreciated.

spi-ca:
Thanks for the quick turnaround, @gotgenes .

I appreciate the investigation and the quick fix.

Issue #385, opened 2026-06-11
pkg:pi-permission-system — Respect pi default active tool set instead of activating all non-denied tools


0xbentang:
### Package

pi-permission-system

### What do you want to change?

`AgentPrepHandler.handle()` in `before-agent-start.ts` uses `pi.getAllTools()` to determine which tools to expose, then calls `pi.setActiveTools()` with every tool that isn't explicitly denied. This activates built-in tools like `find`, `grep`, and `ls` that pi deliberately leaves inactive by default (pi's default active set is `read`, `bash`, `edit`, `write`).

Proposed: start from `pi.getActiveTools()` instead of `pi.getAllTools()` so the permission system only removes tools (deny) or keeps already-active ones — it never activates tools that pi itself left off. This would require adding `getActive()` to the `ToolRegistry` interface and wiring it to `pi.getActiveTools()`.

### Why?

Pi ships `find`, `grep`, and `ls` as built-in tools that are off by default. The subagent system uses these tools in agent definitions via the `tools:` frontmatter (e.g., `tools: read, grep, find, ls`), which controls which tools are passed to `createAgentSession(tools: [...])`. This works because subagents are separate sessions with their own tool set.

But `pi-permission-system`'s `before_agent_start` handler overrides the active tool set for every session — including the main session — by activating all non-denied tools. This means:

1. The main session always sees `find`, `grep`, `ls`, defeating pi's default of keeping them off.
2. Denying these tools globally (`"find": "deny"`) also blocks subagents that explicitly request them, since `before_agent_start` fires for subagent sessions too.
3. The only workaround is per-agent `.md` overrides to re-allow the tools for each subagent type, which is verbose and fragile.

### How?

Two changes:

1. Add `getActive()` to `ToolRegistry` and wire it to `pi.getActiveTools()` in `index.ts`.
2. In `before-agent-start.ts`, replace `this.toolRegistry.getAll()` with `this.toolRegistry.getActive()`.

This way the permission system starts from pi's already-active set and only removes denied tools, preserving pi's default activation decisions.

gotgenes:
Implemented in bf5be48c (breaking) + 79c45944 (interface wiring), shipped in 58db6f816ee31b77deacbb7f613ca3ad692266c7.

**Changes:**
- `feat:` Add `getActive(): string[]` to `ToolRegistry`, wired to `pi.getActiveTools()` in the composition root — all fixtures and fakes updated with `string[]`-typed mocks to match the real SDK shape
[...]

PR #386, opened 2026-06-11
feat(pi-permission-system): respect pi default active tool set

0xbentang:
## Summary

`AgentPrepHandler.handle()` in `before-agent-start.ts` uses `pi.getAllTools()` to determine which tools to expose, then calls `pi.setActiveTools()` with every tool that isn't explicitly denied. This activates built-in tools like `find`, `grep`, and `ls` that pi deliberately leaves inactive by default (pi's default active set is `read`, `bash`, `edit`, `write`).

## Problem

Pi ships `find`, `grep`, and `ls` as built-in tools that are off by default. The subagent system uses these tools in agent definitions via the `tools:` frontmatter (e.g., `tools: read, grep, find, ls`), which controls which tools are passed to `createAgentSession(tools: [...])`. This works because subagents are separate sessions with their own tool set.

But `pi-permission-system`'s `before_agent_start` handler overrides the active tool set for every session — including the main session — by activating all non-denied tools. This means:

1. The main session always sees `find`, `grep`, `ls`, defeating pi's default of keeping them off.
2. Denying these tools globally (`"find": "deny"`) also blocks subagents that explicitly request them, since `before_agent_start` fires for subagent sessions too.
3. The only workaround is per-agent `.md` overrides to re-allow the tools for each subagent type, which is verbose and fragile.

## Proposed change

Use `pi.getActiveTools()` instead of `pi.getAllTools()` as the base set in the `before_agent_start` handler. This way the permission system only removes tools (deny) or keeps already-active ones — it never activates tools that pi itself left off.

Changes:
- Add `getActive()` to `ToolRegistry` interface
- Wire `getActive` to `pi.getActiveTools()` in `index.ts`
- Change `AgentPrepHandler` to call `getActive()` instead of `getAll()`
- Update test mocks to include `getActive`

Ref: #385

### DIFF
--- a/packages/pi-permission-system/src/handlers/before-agent-start.ts
+++ b/packages/pi-permission-system/src/handlers/before-agent-start.ts
@@ -56,7 +56,7 @@ export class AgentPrepHandler {
     this.session.refreshConfig(ctx);
 
     const agentName = this.session.resolveAgentName(ctx, event.systemPrompt);
-    const allTools = this.toolRegistry.getAll();
+    const allTools = this.toolRegistry.getActive();
     const allowedTools: string[] = [];
 
     for (const tool of allTools) {
--- a/packages/pi-permission-system/src/index.ts
+++ b/packages/pi-permission-system/src/index.ts
@@ -152,6 +152,7 @@ export default function piPermissionSystemExtension(pi: ExtensionAPI): void {
 
   const toolRegistry = {
     getAll: () => pi.getAllTools(),
+    getActive: () => pi.getActiveTools(),
     setActive: (names: string[]) => pi.setActiveTools(names),
   };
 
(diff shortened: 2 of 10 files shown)


gotgenes:
@0xbentang Thank you for the report and this PR — it was the basis for the fix that shipped in `pi-permission-system` v11.0.0.

I implemented the change directly on `main` (issue #385) rather than merging this branch, so I could fold in two small improvements while keeping your approach:

- Typed `getActive(): string[]` to match the real `pi.getActiveTools()` contract (instead of `unknown[]`), and updated the mocks/fakes to return bare strings.
- Added a regression test asserting that off-by-default tools (`find`/`grep`/`ls`) present in the registry but absent from the active set are not activated.

Both implementation commits (bf5be48c, 79c45944) carry a `Co-authored-by: Ben Tang <bentang@fastmail.com>` trailer, so the credit lands in the history. Closing in favor of those commits — thanks again!

PR #393, opened 2026-06-12
fix(pi-permission-system): normalize path policy inputs

moekyo:
## Summary

- Add shared cwd-aware policy values for path surfaces so relative inputs can match both legacy relative patterns and absolute allowlists.
- Preserve last-match-wins across equivalent path values, instead of letting a catch-all on the first value mask a later specific rule.
- Make the bash path gate use cd-aware policy values for literal current-shell cd commands while keeping prompts/session approvals on the raw token.
- Document the matching semantics in the README, configuration guide, and JSON schema description.

## Why

This is a focused follow-up to the path-aware extension/MCP work.
After extension and MCP tools started flowing through the same path gates, relative path values such as src/App.jsx could still miss absolute allowlist rules like /workspace/project/* when the current Pi working directory was known.
This keeps the existing flat permission model and path surface semantics, but makes the evaluator consume a more complete set of equivalent path policy values.

## Verification

- CI=true pnpm --filter @gotgenes/pi-permission-system exec vitest run test/input-normalizer.test.ts test/handlers/gates/bash-path.test.ts test/handlers/gates/bash-program.test.ts test/permission-manager-unified.test.ts test/rule.test.ts
- CI=true pnpm --filter @gotgenes/pi-permission-system run check
- CI=true pnpm run lint
- CI=true pnpm run check
- CI=true pnpm run test


gotgenes:
Implemented in 501722b3 — thank you @moekyo for the PR and the clear design analysis!

**Commits (this issue):**
- d34e57fe `feat:` add `normalizePathPolicyLiteral` + `getPathPolicyValues` to `path-utils.ts`
- 2b7d2409 `feat:` add alias-aware `evaluateAnyValue` to `rule.ts`
- 3c2784fc `feat:` thread optional `cwd` through `normalizeInput`; `normalizePathSurfaceValues` returns multi-value aliases
- 6d0c564d **`feat!:`** manager captures `currentCwd`, threads it into `normalizeInput`, dispatches PATH_SURFACES to `evaluateAnyValue`, adds `checkPathPolicy`
- 8ec81da6 `feat:` add `resolvePathPolicy` to `ScopedPermissionResolver` / `PermissionResolver`
- 102a491e `feat:` add `pathRuleCandidates(cwd?)` to `BashProgram` — cd-aware policy values per token
- 7bcdbe70 **`feat!:`** bash path gate migrates to `pathRuleCandidates` + `resolver.resolvePathPolicy`
- 49ba84d3 `refactor:` remove orphaned `BashProgram.pathTokens` / `extractTokensForPathRules` chain
- 8ab53a2d `docs:` document cwd-aware matching in README, configuration guide, and JSON schema

**Behavior change:** When Pi's working directory is known, a relative path input (tool `input.path`, bash token) now also evaluates against its cwd-normalized absolute form. A config like `path: { "*": "ask", "/workspace/project/*": "allow" }` will move a relative `src/App.jsx` from `ask` → `allow`. Bash tokens resolve against the effective directory after literal `cd` commands; tokens after a non-literal `cd "$DIR"` stay conservative (literal form only).

**Design note:** The simplified design drops the `INTERNAL_PATH_POLICY_VALUES` symbol side-channel from the PR in favour of an explicit `checkPathPolicy` / `resolvePathPolicy` method pair — policy values are computed internally by the bash gate and never read from a user-controlled field on tool input.

⚠️ **Breaking change** (two `feat!:` commits, `BREAKING CHANGE:` footers in each). Users with absolute allowlist rules (`/abs/path/*`) will now also match relative inputs that resolve to the same path. To preserve stricter control, narrow patterns or add an explicit `path` deny for sensitive paths.

PR #395, opened 2026-06-12
feat(pi-permission-system): deny patterns with custom reason

k0valik:
> slop from my agent incoming, should be clean tho

## Summary

Extend permission patterns to support an object syntax for deny actions with an optional custom reason.

### Before

```json
"bash": { "npm *": "deny" }
```

### After

```json
"bash": { "npm *": { "action": "deny", "reason": "Use pnpm instead" } }
```

The reason is shown to the agent when the command is blocked:

```
[pi-permission-system] is not permitted to run 'bash' command 'npm install' (matched 'npm *'). Reason: Use pnpm instead.
```

### Changes

- **Types:** `DenyWithReason` interface, `PatternValue` type, `reason` on `PermissionCheckResult`
- **Rule:** `reason?: string` on `Rule`
- **Normalizer:** Parse `{ action: "deny", reason }` objects in `normalizeFlatConfig`
- **Config loader:** Accept `DenyWithReason` from JSON config files (gap in the design doc, was silently stripped)
- **Denial messages:** Append reason after the sentence-ending period
- **Permission manager:** Thread `rule.reason` into result
- **Schema:** Extended `permissionState` with `{ action, reason }` object form
- **Config + docs:** Updated example config and configuration docs
- **Tests:** 17 new tests across normalize, rule, denial-messages, and permission-manager-unified suites

### Verification

- ✅ 1989 tests pass (94 files, 0 regressions)
- ✅ TypeScript `tsc --noEmit` clean
- ✅ Biome lint clean on all changed files
- ✅ rumdl clean on docs
- ✅ All existing configs continue to work unchanged (backward compatible)

### DIFF
--- a/packages/pi-permission-system/config/config.example.json
+++ b/packages/pi-permission-system/config/config.example.json
@@ -25,7 +25,8 @@
       "*": "ask",
       "git *": "ask",
       "git status": "allow",
-      "git diff": "allow"
+      "git diff": "allow",
+      "npm *": { "action": "deny", "reason": "Use pnpm instead" }
     },
     "mcp": { "*": "ask", "mcp_status": "allow", "mcp_list": "allow" },
     "skill": { "*": "ask" },
--- a/packages/pi-permission-system/docs/configuration.md
+++ b/packages/pi-permission-system/docs/configuration.md
@@ -58,7 +58,13 @@ Scalar fields (`debugLog`, `permissionReviewLog`, `yoloMode`) use simple replace
     "read": "allow",
     "write": "deny",
     "edit": "deny",
-    "bash": { "git *": "ask", "git status": "allow" },
+    "bash": {
+      "*": "ask",
+      "git *": "ask",
+      "git status": "allow",
+      "git diff": "allow",
+      "npm *": { "action": "deny", "reason": "Use pnpm instead" }
+    },
     "mcp": { "mcp_status": "allow" },
     "skill": { "*": "ask" },
     "external_directory": "ask"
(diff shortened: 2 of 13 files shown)


gotgenes:
Implemented in 51750e18 through d8e57563 (with a docs/refactor cleanup in fd880d49 and 4c2dc7c9).

Thanks to @k0valik for the original PR — the capability and API shape came directly from that work, and every implementation commit carries a `Co-authored-by` trailer crediting the contribution.

**Commits:**
- 51750e18 — add `DenyWithReason` type and shared `isDenyWithReason` guard in `common.ts`
- 186c15a7 — build deny rules with reason in `normalizeFlatConfig`
- 3201bfd5 — preserve deny-with-reason objects through the JSON config loader (previously stripped silently)
- ed712e47 — thread `reason` into `PermissionCheckResult` via `buildCheckResult`
- d8e57563 — append `Reason: …` to agent-facing denial messages

**Behavior:** deny rules now accept an optional object form at the pattern-value level:

```json
"bash": { "npm *": { "action": "deny", "reason": "Use pnpm instead" } }
```

When that pattern matches, the agent sees the reason appended to the block message:

```
[pi-permission-system] is not permitted to run 'bash' command 'npm install' (matched 'npm *'). Reason: Use pnpm instead.
```

Non-breaking — existing string-form config is unchanged. The `reason` field is optional and only valid on `deny` rules.

**Design simplifications over the PR:**
- Single shared `isDenyWithReason` guard in `common.ts` (PR had two divergent copies)
- `DenyWithReason` is valid only at the pattern-value level, not at the surface level (type and schema match runtime behaviour)
- JSON schema `$defs/denyWithReason` is referenced only from `permissionMap`, never the surface-level `oneOf`

PR #435, opened 2026-06-19
fix(pi-permission-system): strip shell comment lines from bash commands before matching

rnavarro:
## Symptom

I use wildcard "always allow" patterns and session-approval suggestions for bash commands. When an agent prepends a descriptive comment line before the actual command (a common pattern, e.g. `# list node versions` then `nvm ls`), the pattern match and the suggestion both fired against the comment text instead of the command, so an approved `nvm ls` pattern did not match and the suggestion proposed a pattern built from the `# list...` tokens.

## What I observed

Both `normalizeInput` (bash surface lookup value) and `suggestBashPattern` tokenize the leading text of the command. A leading `#`-comment line shifts the "leading text" onto the comment, so the lookup and the suggestion operate on comment tokens rather than the command.

## Change

Add `stripBashCommentLines()` to `bash-arity.ts` (the shared module already responsible for bash command structure analysis) and apply it in both consumers:

- `normalizeInput` (bash surface): strip comments for the lookup value, falling back to the raw command when nothing meaningful remains, so an all-comment command still evaluates against its literal text.
- `suggestBashPattern`: strip comments before tokenizing so suggestions are based on the real command.

The original command is preserved in `resultExtras` for display and logging.

## Tests

`test/bash-arity.test.ts`, `test/input-normalizer.test.ts`, and `test/pattern-suggest.test.ts` cover the stripping, the fallback for all-comment commands, and that suggestions are computed against the command rather than the comment.

### DIFF
--- a/packages/pi-permission-system/src/bash-arity.ts
+++ b/packages/pi-permission-system/src/bash-arity.ts
@@ -184,3 +184,27 @@ export function prefix(tokens: string[]): string[] {
   // Unknown command — default arity 1.
   return [tokens[0]];
 }
+
+/**
+ * Remove shell comment lines from a bash command string.
+ *
+ * A comment line is one whose first non-whitespace character is `#`. Agents
+ * frequently prepend descriptive comments before the real command
+ * (e.g. `"# Check debug logs\nfind ..."`); such prefixes defeat wildcard
+ * pattern matching and session-approval suggestions, which tokenize the
+ * leading text. Stripping comment lines lets matching operate on the actual
+ * command.
+ *
+ * The original command is never returned: when every line is a comment (or
+ * the input is blank) an empty string is returned, and each caller applies
+ * its own fallback.
+ *
+ * @param command - Raw bash command, possibly multi-line.
+ * @returns The command with comment lines removed and surrounding whitespace
+ *   trimmed, or an empty string when nothing meaningful remains.
+ */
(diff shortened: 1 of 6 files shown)


gotgenes:
@rnavarro Taking a look now. Thanks for submitting!

gotgenes:
Merged via rebase as `d0455915` — thank you, @rnavarro! 🙏

This is a clean, well-scoped fix. A few things I appreciated on review:

- `stripBashCommentLines()` lands in `bash-arity.ts` alongside `ARITY`/`prefix`, exported-for-testability in the same style — good convention fit.
- The `/^\s*#/` filter is conservative: it only strips lines whose first non-whitespace char is `#`, so `echo "#foo"` and inline trailing comments stay intact. Gating now operates on the *real* command instead of a comment-confused value, which is security-positive.
- Preserving the original in `resultExtras.command` (with the `|| command` fallback for all-comment input) keeps display/logging and the all-comment edge case correct.

Verified on the branch: `tsc` clean, lint clean, 2033 tests pass, `fallow dead-code` clean. Classified as a non-breaking `fix:` — it makes a previously comment-defeated match align with a pattern the user explicitly approved, without loosening policy (`# nvm ls\nrm -rf /` still strips to `rm -rf /` and won't match an `nvm ls` pattern).

Appreciate the thorough tests across all three consumers.

Issue #443, opened 2026-06-20
pi-subagents: consolidate remaining test clone families


gotgenes:
## What

Consolidate the remaining test clone families surfaced by fallow, run after the Phase 19 terminal cut so no helper is extracted into a file the cut then deletes.

## Why

fallow reports 16 test clone groups at the Phase 18 end.
The terminal cut (Phase 19 Steps 5–6) deletes several test files and removes roughly 4 of those groups automatically; the survivors are extraction targets.
Running this last means we only consolidate clones that actually survive.

## Proposed change

- `test/lifecycle/subagent-manager.test.ts` — extract a shared assertion helper for 3 clone families (23 lines across groups at :92/:109, :282/:330, and :323 shared with `subagent.test.ts`).
- `test/ui/agent-widget.test.ts` — merge the duplicate `makeWidget` helper defined twice across two `describe` blocks (14-line clone at :225/:284).
- `test/session/session-config.test.ts` — extract a shared fixture for the 16-line internal clone (lines 131–146 / 151–166).
- `test/lifecycle/concurrency-limiter.test.ts` — extract shared setup for the 10-line clone (lines 21–30 / 148–155).
- `test/tools/spawn-config.test.ts` — extract a shared fixture for the 9-line clone (lines 22–30 / 35–43).

Target: test clone groups ≤ 10 (from 16).

## Context

Phase 19 roadmap, Track C (test health).
Best run after Phase 19 Steps 5 and 6 land.
See the Phase 19 section in `packages/pi-subagents/docs/architecture/architecture.md`.
Independently releasable.

gotgenes:
Implemented in 693e26b4474d44091610b4293b9813b7144263e1 (through faa98886a7df08d1d3dcc6e379d1f963737e6048).

Commits:
- `test(pi-subagents): extract shared makeNavigable test fixture`
- `test(pi-subagents): hoist duplicated makeWidget helper`
- `test(pi-subagents): extract captured-overlay render helper`
- `test(pi-subagents): consolidate subagent-manager arrange helpers`
- `test(pi-subagents): consolidate subagent ready-state arrange`
- `test(pi-subagents): extract preparedBracket setup helper`
- `docs(pi-subagents): reconcile Phase 19 Step 7 outcome with achieved clone count`

[...]

Issue #448, opened 2026-06-20
`enabled: false` does not prevent explicitly spawning disabled agents


nickadminroot:
### Package

pi-subagents

### Package version

v17.0.0

## Summary

In `@gotgenes/pi-subagents` v17.0.0, setting `enabled: false` in a custom agent override does not actually prevent that agent from being spawned when it is referenced explicitly via `subagent_type`.

This appears to contradict the README and the inline registry comment, which describe disabled agents as excluded from spawning.

I searched existing open/closed issues and PRs for `enabled false`, `disabled agent`, `resolveType`, and `subagent_type`, but did not find an existing report for this behavior.

## Expected behavior

The disabled agent should not be spawned.

Possible acceptable behaviors:

1. Treat the disabled type as unavailable/unknown and fall back to `general-purpose`, matching the existing unknown-type fallback behavior.
2. Return an explicit error such as `Agent type "Plan" is disabled`.

Either behavior would be consistent with `enabled: false` disabling the agent for spawning.

## Actual behavior

The disabled agent is still resolved and spawned.

It is removed from `getAvailableTypes()` and therefore from the `availableTypesText` part of the tool parameter description, but an explicit `subagent_type: "Plan"` still works.

## Suspected cause

In `src/config/agent-types.ts`, `resolveType()` delegates to `resolveKey()`:

```ts
resolveType(name: string): string | undefined {
  return this.resolveKey(name);
}
```

But `resolveKey()` only checks whether the key exists in the registry. It does not check `enabled`:

```ts
private resolveKey(name: string): string | undefined {
  if (this.agents.has(name)) return name;
  const lower = name.toLowerCase();
  for (const key of this.agents.keys()) {
    if (key.toLowerCase() === lower) return key;
  }
  return undefined;
}
```

The spawn path then accepts that result:

```ts
const rawType = params.subagent_type as SubagentType;
const resolved = registry.resolveType(rawType);
const subagentType = resolved ?? "general-purpose";

const customConfig = registry.resolveAgentConfig(subagentType);
const resolvedConfig = resolveAgentInvocationConfig(customConfig, params);
```

`resolveAgentConfig()` also returns the config for an existing disabled key without checking `enabled`.

There is an `isValidType()` method that checks `enabled`, but it does not appear to be used in the spawn resolution path.

### Steps to reproduce

Create a project-level override for a default agent, for example:

`.pi/agents/Plan.md`

```md
---
enabled: false
description: Disabled Plan agent
[...]

PR #594, opened 2026-07-15
fix(pi-subagents): complete exclude disabled agents from the subagent…

whaoa:
## Summary

Fixes the pi-subagents tool description so disabled built-in agents are fully omitted from the exposed subagent tool help.

- omits the entire “Default agents” section when no default agents are available.
- removes hard-coded usage guidelines for disabled built-in agents such as Explore, Plan, and general-purpose.
- adds regression coverage for disabled built-in agents and empty default-agent lists.

## Why

Previously, disabled built-in agents could still appear in the subagent tool description through static guideline text, even after being excluded from the available agent list. The change makes the generated description reflect the actual registry state, keeping model guidance consistent with the enabled agent surface.

This problem relates to Issue #443 which has already been resolved, but the fix in commit 9a43414b3e15f0c978db9d468b737b13b801fdb2 is incomplete.

## Verification

- CI=true pnpm run --filter @gotgenes/pi-subagents exec vitest run test/helpers/make-deps.ts test/tools/agent-tool.test.ts test/tools/helpers.test.ts
- CI=true pnpm --filter @gotgenes/pi-subagents run check
- CI=true pnpm --filter @gotgenes/pi-subagents run lint

### DIFF
--- a/packages/pi-subagents/src/tools/agent-tool.ts
+++ b/packages/pi-subagents/src/tools/agent-tool.ts
@@ -128,6 +128,12 @@ export class AgentTool {
 		const agentDir = this.agentDir;
 		const registry = this.registry;
 
+		const agentSpecificGuidelines = [
+			registry.isValidType("Explore") &&  "- Use Explore for codebase searches and code understanding.",
+			registry.isValidType("Plan") &&  "- Use Plan for architecture and implementation planning.",
+			registry.isValidType("general-purpose") && "- Use general-purpose for complex tasks that need file editing.",
+		].filter((line): line is string => typeof line === "string");
+
 		return defineTool({
 			name: "subagent" as const,
 			label: "Subagent",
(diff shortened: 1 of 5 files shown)


gotgenes:
Thanks for the report and the PR, @whaoa — the underlying gap was real: #448 excluded disabled built-in agents from the subagent tool's type list, but left three hardcoded `Guidelines:` lines (`Explore`, `Plan`, `general-purpose`) that still referenced those agents even when disabled.

Implemented in a2b41a665586a271b6453b89dabec0bd94be394c and a29c32498caa4e29f0cc3292d5a6be4f4126c5f8, adopting your PR's capability with a simplified design (recorded in `packages/pi-subagents/docs/retro/0594-complete-exclude-disabled-agents-tool-description.md`):

- `fix(pi-subagents): source subagent guideline copy from agent config` — added a `toolGuideline` field to `AgentConfig`, populated it on the three embedded defaults, and sourced the `Guidelines:` block from the registry instead of hardcoding the lines, so a disabled built-in now drops its guideline automatically.
- `fix(pi-subagents): omit empty Default agents section header` — `buildTypeListText` now omits the bare `Default agents:` header when no default agents are enabled, matching your PR's second fix.

User-visible behavior: when a built-in agent is disabled via `enabled: false`, the subagent tool description no longer references it anywhere — not the type list (already fixed in #448) and now not the `Guidelines:` block either. Non-breaking for the default (all-enabled) configuration; the only change there is that the three guideline lines now appear in registry order (general-purpose, Explore, Plan) rather than their previous hardcoded order — same text, reordered.

Your PR's test coverage approach was preserved (disabled-builtin regression tests, empty-default-list coverage) with corrected assertions — thanks again for surfacing this.

PR #612, opened 2026-07-17
feat(pi-subagents): inherit parent active tools

wolfgangmeyers:
## Summary
- capture the parent active tool set at the extension runtime composition boundary
- persist it in the immutable spawn snapshot
- include active extension-provided parent tools in the native child’s initial tool set before its first model/provider request
- enable tools such as Mailbox `mecha_parent_message` to be available when the provider is loaded in that child
- preserve native child completion and the existing recursion guard after extensions bind
- add no relay, Sentinel-specific path, or policy layer

## Validation
- `mise exec pnpm@11.5.2 -- pnpm --filter @gotgenes/pi-subagents run check`
- `mise exec pnpm@11.5.2 -- pnpm --filter @gotgenes/pi-subagents test` (64 files, 999 tests)
- independent final review: clean

### DIFF
--- a/packages/pi-subagents/src/index.ts
+++ b/packages/pi-subagents/src/index.ts
@@ -54,6 +54,7 @@ export default function (pi: ExtensionAPI) {
 
   // ---- Runtime: all mutable extension state in one place ----
   const runtime = createSubagentRuntime();
+  runtime.setActiveToolsProvider(() => pi.getActiveTools());
 
   // ---- Notification system ----
   // Owns completion nudges and live-activity cleanup. The widget detects finished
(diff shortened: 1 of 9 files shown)


gotgenes:
Thanks for this, @wolfgangmeyers — same underlying problem as #725, and reading this PR's approach was genuinely useful while diagnosing it.

Closing as superseded: #725 landed a different fix, and this approach has two issues worth naming for anyone who finds this PR later.

1. `tools: [...cfg.toolNames, ...snapshot.activeTools]` unions in the **parent's whole active set**, which includes built-ins. A read-only `Explore`/`Plan` agent (`tools: read, bash, grep, find, ls` — no `edit`/`write`) would gain `edit` and `write` the moment the parent has them active, since nothing here filters `snapshot.activeTools` down to extension-registered names.
2. Admitting `subagent` into a child's allowlist (present in the parent's active set) reopens the recursion guard: `applyRecursionGuard`'s post-bind filter runs once, but Pi's `_refreshToolRegistry` re-adds every allowlisted registry tool to the active set on each subsequent refresh — including the refresh triggered by a child extension's own `registerTool()` call. The guard's effect doesn't survive the next child-side registration.

#725 keeps the default (a child's tools are exactly its agent's `tools:` list) and instead documents extension tool names as a supported `tools:` entry, fixes the YAML-sequence parsing, and replaces the post-bind guard with an SDK-level `excludeTools` denylist that's reapplied on every refresh — closing exactly the hole in point 2 above.

PR #613, opened 2026-07-18
feat(pi-subagents): add global background default

wolfgangmeyers:
## Summary

Adds the documented `defaultRunInBackground` global setting in `~/.pi/agent/subagents.json`.
When set to `true`, it makes root `subagent` calls that omit `run_in_background` launch in the background.

Explicit `run_in_background: false` still launches in the foreground.
Explicit agent-frontmatter behavior continues to win over the global fallback.
Nested child calls remain foreground by default, and forked root sessions still receive the global default.
The service adapter is unchanged.

## Validation

- `mise exec pnpm@11.5.2 -- pnpm --filter @gotgenes/pi-subagents run check`
- `mise exec pnpm@11.5.2 -- pnpm --filter @gotgenes/pi-subagents test` — 64 files, 1007 tests passed
- `mise exec pnpm@11.5.2 -- pnpm --filter @gotgenes/pi-subagents run lint`
- `git diff --check`

Numeric coverage is not available in this package because it has no coverage script or Vitest coverage provider, and `@vitest/coverage-v8` is not installed.
The focused behavioral tests cover the global true/false defaults, explicit foreground override, agent-frontmatter precedence, nested-child non-regression, and forked-root behavior.

### DIFF
--- a/packages/pi-subagents/README.md
+++ b/packages/pi-subagents/README.md
@@ -266,6 +266,8 @@ Foreground agents bypass the queue — they block the parent anyway.
 ## Persistent Settings
 
 Runtime tuning values set via `/subagents:settings` (max concurrency, default max turns, grace turns) persist across pi restarts.
+The hand-edited global config also supports `defaultRunInBackground` (default `false`): when `true`, any `subagent` call that omits `run_in_background` runs in the background.
+Explicit `run_in_background: true` or `false` and agent frontmatter continue to take precedence.
 Two files, merged on load:
 
 - **Global:** `~/.pi/agent/subagents.json` — your machine-wide defaults.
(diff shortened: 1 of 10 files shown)


gotgenes:
Thanks for the PR, and I apologize for how long it took me to answer it.

I'm not going to take this one, and the reason is a scope boundary rather than anything about the implementation — which is careful. The precedence handling (explicit argument, then frontmatter, then global fallback) is exactly right if the setting existed.

A global run-mode default is a stated non-goal for this package:

> _A global run-mode default._
> Foreground or background is a per-invocation argument and a per-agent frontmatter key; a global flip changes every existing agent file at once.

That is the part I can't get past. `defaultRunInBackground: true` changes how every agent file already on disk behaves, including ones whose authors chose foreground by writing nothing at all. The two existing mechanisms — `run_in_background` on the call and the frontmatter key on the agent — both say what they mean at the place someone reads them.

The friction you're solving is real. If the problem is that *your* agents should default to background, the per-agent frontmatter key does that today without reaching across anyone else's agent files.

#615 is a separate question; I'll answer it on its own thread.

Issue #604, opened 2026-07-16
pi-permission-system: option to widen session-approval pattern


AndersBennedsgaard:
### Package

pi-permission-system

### What do you want to change?

Some way to approve a broader scope for the session.

For example a config know to control suggestion granularity (for example repo root or N levels up from the file) or make the existing dialog option editable.

I would probably suggest a config change, such that you can set:

```jsonc
{
  "sessionApprovalScope": "parent-dir" | "repo-root" | number // levels up from file
}
```

Default stays `"parent-dir"` (current behavior).

### Why?

When an `ask` decision fires for a path-bearing tool (`write`/`edit`/etc.), the "Yes, allow ... for this session" dialog option suggests a wildcard pattern derived by `deriveApprovalPattern()`

https://github.com/gotgenes/pi-packages/blob/c79302d5fa440ac54ecd304431087084bb5f79a7/packages/pi-permission-system/src/session-rules.ts#L67-L79

This always scopes the suggested pattern to the file's **immediate parent directory**, with no way to widen it. For a deeply nested file (e.g. `repo/packages/foo/src/deep/nested/module.ts`), the suggested session approval only covers `repo/packages/foo/src/deep/*`. So working across a whole repo (or even a whole package) means re-approving the same "for this session" prompt over and over for every new subdirectory the agent touches.

Issue #603, opened 2026-07-16
Allow permission for current project


LoricAndre:
### Package

pi-permission-system

### What do you want to change?

Some permissions should not be automatically allowed for the system but might be in the project. It would be nice to be able to do this from the prompt, adding a new `p` shortcut for example that edits the project permissions file.

### Why?

To avoid authorizing commands and breaking the flow on every session of a project

PR #638, opened 2026-07-23
feat(pi-permission-system): add overlay config option

FabianUntermoser:
## Problem

The inline permission dialog (`{ overlay: false }`) replaces the full TUI content on every prompt. In tmux, this causes visible flicker on scroll.

## Fix

Add `overlay` config option. When `true`, the dialog renders on top of existing content - only the overlay portion redraws. Default `false` (existing behavior).

### DIFF
--- a/packages/pi-permission-system/src/authority/local-user-authorizer.ts
+++ b/packages/pi-permission-system/src/authority/local-user-authorizer.ts
@@ -55,6 +55,7 @@ export class LocalUserAuthorizer implements TerminalAuthorizer {
         ui: this.deps.ui,
         doublePressToConfirm:
           this.deps.getPromptPreferences().doublePressToConfirm,
+        overlay: this.deps.getPromptPreferences().overlay,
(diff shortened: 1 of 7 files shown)


gotgenes:

**It's parked, and on what.** Adding a configuration channel for permission policy is a design decision rather than a code review. Which channels policy may enter through, and with what precedence and trust gating, is being settled in #799. Several open requests — #603, #604, #691, #720, and PRs #671, #675, #684, #692 — are all asking for related widenings, and I'd rather answer them once from a written decision than one at a time and end up with an inconsistent surface.

**The tree moved underneath your branch.** #837 reorganized `pi-permission-system/src`, moving the 42% of files that sat at the package root into subdirectories. Five of the seven `src` files this PR touches were moved, so rebasing isn't worth your time right now — once #799 is decided this will mean reapplying the change against the new layout rather than resolving conflicts.

Thanks for your patience.

Issue #642, opened 2026-07-24
pi-permission-system: preserve Ctrl+O tool expansion in inline permission prompts


0xbentang:
### Package

pi-permission-system

### Package version

20.10.0

### What happened?

While an inline permission prompt is focused, Ctrl+O (`app.tools.expand`) has no effect. Tool rows cannot be expanded or collapsed while reviewing a pending tool call.

The custom permission component handles its decision and navigation keys but ignores the injected keybindings manager. Its UI surface also excludes `getToolsExpanded()` and `setToolsExpanded()`.

This makes it harder to inspect the complete pending tool invocation before approving it.

### Expected behavior

Ctrl+O should toggle tool expansion while the permission prompt remains open, matching Pi's standard `ExtensionSelectorComponent`. It should not resolve or otherwise alter the permission decision.

### Steps to reproduce

1. Configure a safe command such as `curl *` with permission `ask`.
2. Run `curl --version`.
3. While the permission prompt is open, press Ctrl+O.
4. Observe that the tool display does not expand or collapse.

### Additional context

Pi 0.81.1 on macOS 26.5.1.

The inline prompt uses `ctx.ui.custom(..., { overlay: false })` and receives keyboard focus. A focused custom component must explicitly preserve application actions such as `app.tools.expand`.

A narrow fix could match `app.tools.expand` using the injected keybindings manager and toggle `ui.getToolsExpanded()` through `ui.setToolsExpanded()`, with a test confirming that the permission decision remains unchanged.

gotgenes:
Implemented in 6a0d241291f8eefa51493b683e12514a83f295bd (docs follow-up in f4098d33).

Thanks to @0xbentang for the report and for PR #643, which surfaced the gap and worked out the fix against Pi's `KeybindingsManager`/`ExtensionUIContext` surface. We took a slightly narrower path for the fix itself (a single `(data: string) => boolean` seam instead of threading the keybindings manager into the component), but the capability and the underlying diagnosis are entirely his.

**What changed:** while the inline TUI permission prompt holds keyboard focus, Pi's `app.tools.expand` action (Ctrl+O by default) now toggles tool-output expansion, so you can expand a truncated tool preview before approving or denying. The toggle only affects display — it never resolves, commits, or arms the pending decision — and it is not intercepted while typing a denial reason, so a rebound key still reaches the reason editor. Non-TUI sessions (RPC/frontend-driven) are unaffected.

[...]

PR #643, opened 2026-07-24
fix(pi-permission-system): preserve tool expansion in prompts

0xbentang:
## Summary

- preserve Pi's configured `app.tools.expand` action while the inline permission prompt has focus
- toggle native tool expansion without changing or resolving the pending permission decision
- add regression coverage for expansion, collapse, and continued prompt interaction

## Verification

- `pnpm --filter @gotgenes/pi-permission-system test`
- `pnpm --filter @gotgenes/pi-permission-system lint`

Refs #642

### DIFF
--- a/packages/pi-permission-system/src/authority/permission-prompt-component.ts
+++ b/packages/pi-permission-system/src/authority/permission-prompt-component.ts
@@ -1,6 +1,7 @@
 import type {
   ExtensionContext,
   ExtensionUIContext,
+  KeybindingsManager,
 } from "@earendil-works/pi-coding-agent";
 import {
   type Component,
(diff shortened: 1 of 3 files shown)


gotgenes:
Thanks for filing this, @0xbentang — the diagnosis (the inline permission prompt's `ctx.ui.custom` component swallows every keystroke, so `app.tools.expand` was dead while it held focus) and the fix approach here were both correct, and your PR was the reference we implemented against.

We landed a simplified version of the same capability in 6a0d241291f8eefa51493b683e12514a83f295bd (docs in f4098d33), closing #642. The main differences from this PR:

- The prompt component takes one narrow `(data: string) => boolean` app-action seam instead of threading the full `KeybindingsManager` in — keeps the SDK type out of the component and needs no keybindings fake in tests.
- The action is consulted only in the decision/scope steps, explicitly guarded against being intercepted while a denial reason is being typed.
- Docs in `README.md` and `docs/configuration.md` note the behavior.

Closing this as superseded by that implementation, but the credit for finding and diagnosing the gap is yours — thank you.

Issue #654, opened 2026-07-25
pi-permission-system: add contextual natural-language explanations to ask prompts


marcoscale98:
### Package

- pi-permission-system
- pi-permission-model-judge

### What do you want to change?

Add an optional, model-assisted explanation to interactive `ask` prompts. Before the human chooses whether to allow or deny a tool call, a configured lightweight model should explain the requested operation in plain language using the relevant context (similar to Claude Code).

The goal is explanation, not automated advice or adjudication. For example, a bash prompt could show:

> This command downloads the remote branch, rebases your local commits on top of it, and may rewrite local commit hashes. It contacts the configured Git remote but does not push any changes.

The explanation should appear in the existing permission dialog above the normal choices (`Yes`, `Yes, for this session`, `No`, and `No, provide reason`). The human remains the terminal authority and the explainer must not return `allow` or `deny`, suppress the prompt, or recommend a choice.

Useful context may include the permission surface, command or tool input, cwd, agent name, the user's current request, and a small bounded slice of recent conversation. The extension should make the included context explicit and configurable so users can balance explanation quality, latency, cost, and privacy.

### Why?

Permission prompts often contain commands that are difficult to evaluate without mentally expanding flags, shell composition, filesystem effects, network access, and the intent of the current task. A raw command such as `git pull --rebase`, `find ... -exec ...`, or a package-manager invocation may be familiar in isolation but have different implications depending on cwd and the work the agent is currently performing.

The current model-judge and Authorizer chain solve a different problem: they make case-by-case `allow` / `deny` / `defer` decisions. The first-party model judge is intentionally deny-first and limited to typo-path `external_directory` asks. I want a human-facing explanation that always defers to the existing interactive decision.

This would help users make informed decisions without weakening deterministic policy or delegating authority to an LLM. If the model is unavailable, times out, or produces an invalid response, the current permission prompt should appear unchanged.

Related work includes #472, #591, #599, #600, and #620, but none appears to cover a contextual, non-decisional explanation rendered inside the human prompt.

### How? (optional)

One possible design is an explicit advisory seam separate from authorization, for example a registered `PermissionPromptExplainer` that receives a read-only, privacy-bounded projection of `PromptPermissionDetails` plus selected session context and returns structured display content such as:

```ts
{
  summary: string;
  effects?: string[];
  risks?: string[];
}
```

[...]

Issue #648, opened 2026-07-24
pi-permission-system: show edit diffs before permission approval


marcoscale98:
### Package

pi-permission-system

### What do you want to change?

Show a human-readable diff for pending `edit` tool calls before the user approves an `ask` permission prompt.

In version 20.10.0, the edit formatter only presents the target path and a count-based summary such as “1 replacement: edit #1 replaces 1 line with 1 line.” That confirms the size of the operation, but not what text will actually be removed or inserted.

The approval UI should let the user inspect the complete proposed change as a conventional diff while the permission decision is still pending. A compact summary can remain the default view, with the diff available directly or through an explicit “View diff” action.

Related: #642 requests preserving Ctrl+O tool expansion while an inline permission prompt is focused. Even if tool expansion is restored, a dedicated edit diff would make the security-relevant content substantially easier to review than raw tool arguments.

### Why?

A permission prompt for file modification is useful only if the user can make an informed decision. The path and changed-line counts are insufficient when approving edits to sensitive files such as `AGENTS.md`, CI configuration, security policy, or credentials-related configuration.

Without a diff, the user must approve based on incomplete information or deny the call and inspect it through another workflow. This creates pressure to approve blindly and weakens the value of the permission gate.

A pre-execution diff would provide an experience closer to Claude Code’s edit approval flow and make `ask` meaningfully reviewable.

### How? (optional)

- For `edit`, render each `oldText` / `newText` pair as a bounded unified diff before resolving the permission decision.
- Keep the current path/count summary as the collapsed view.
- Add a “View diff” action or integrate with `app.tools.expand` after #642 is addressed.
- Ensure long replacements are truncated safely, with a clear indication that the preview is incomplete.
- Do not write the file or otherwise execute the tool to produce the preview.

Current count-only formatter:
https://github.com/gotgenes/pi-packages/blob/1a904d234b769f51f4872b8e2b692a371cb02d69/packages/pi-permission-system/src/tool-input-prompt-formatters.ts#L8-L39

gotgenes:
Thanks for this, and I apologize for the long silence on it — the answer took a while to become sayable, and then I did not come back and say it.

The short version: I agree with the need, it belongs downstream rather than in this package, and the seam you would build it on already ships.

## Why not in this package

[...]

Issue #519, opened 2026-07-01
Clarify SDK custom UIContext support for frontend-driven permission approvals


navlisData:
### Package

pi-permission-system

### What do you want to change?

What do you want to change?

Please clarify and document the intended SDK integration path for frontend-driven permission approvals in `@gotgenes/pi-permission-system`.

I am embedding Pi via the [SDK](https://github.com/earendil-works/pi/tree/main/packages/coding-agent) and want permission `ask` prompts to be rendered in my own frontend. After reading the source, it looks like this may already be supported through Pi’s normal `ExtensionUIContext` abstraction:

* SDK app creates an `AgentSession`
* SDK app calls `session.bindExtensions({ uiContext, mode: ... })`
* `pi-permission-system` emits `permissions:ui_prompt` for observability
* then it awaits the active UI path
* when `ctx.hasUI` is true, `PermissionForwarder` calls `requestPermissionDecisionFromUi(ctx.ui, ...)`
* `requestPermissionDecisionFromUi` uses `ctx.ui.select(...)` and optionally `ctx.ui.input(...)`
* the selected value resolves the blocked permission request and the agent/subagent continues

Could you confirm whether this is the intended and supported integration path for SDK embeddings with custom frontends?

In particular:

* Is `permissions:ui_prompt` intentionally observability-only?
* Is there intentionally no `permissions:ui_prompt:reply:<requestId>` approval channel?
* Should SDK integrations provide their own `ExtensionUIContext` and resolve permission decisions through `select(...)` / `input(...)`?
* For `@gotgenes/pi-subagents`, should forwarded child `ask` prompts also resolve through the parent session’s custom `uiContext`?
* Are there any constraints on the `mode` value SDK apps should pass when binding a custom non-terminal frontend UI?

### Why?

For SDK-based integrations, the host application often already has its own web, desktop, or IDE UI. In that setup, permissions should be shown in the host frontend rather than in Pi’s built-in terminal UI or through JSONL RPC.

The source suggests that this may already work via `ExtensionUIContext`, which would be a clean and intentional design rather than a package-specific permission reply API. However, from the current documentation it is not immediately clear whether frontend-driven approvals should be implemented this way or whether `permissions:ui_prompt` / `permissions:rpc:prompt` are expected to be involved.

Documenting the recommended pattern would help SDK embedders avoid relying on private internals, avoid treating `permissions:ui_prompt` as a replyable request, and correctly support subagent permission forwarding.

[...]

PR #656, opened 2026-07-25
fix(pi-permission-system): cap permission prompt width to prevent TUI crash

ydeng11:
## Problem

Permission prompts can produce lines wider than the terminal (e.g., 279 visible width vs 213 terminal columns), triggering a TUI crash in `pi-tui` (`@earendil-works/pi-tui`). This happens because `formatAskPrompt()` assembles a message like:

```
Current agent requested tool 'ctx_batch_execute' with input {long JSON...}. Allow this call?
```

The `toolInputPreviewMaxLength` (default 200) only limits the JSON portion, but the prefix (~58 chars) and suffix (~17 chars) push the total well past terminal width.

## Fix

Import `truncateInlineText` from `./tool-input-preview` and wrap each branch of `formatAskPrompt()` with a `MAX_PROMPT_WIDTH` of **200 characters** — applied to the **final assembled message**, not just the JSON portion.

Also added the truncation to the bash command and MCP branches for defense-in-depth.

## Testing

All **2593 existing tests pass** (including the 29 `permission-prompts.test.ts` cases).

## Related

This pairs with a companion fix in `@earendil-works/pi-tui` that replaces the crash-with-throw with safe truncation:
- earendil-works/pi#7116

### DIFF
--- a/packages/pi-permission-system/src/permission-prompts.ts
+++ b/packages/pi-permission-system/src/permission-prompts.ts
@@ -4,6 +4,10 @@ import type { SkillPromptEntry } from "./skill-prompt-sanitizer";
 import type { ToolPreviewFormatter } from "./tool-preview-formatter";
 import type { PermissionCheckResult } from "./types";
 import { getNonEmptyString, toRecord } from "./value-guards";
+import { truncateInlineText } from "./tool-input-preview";
+
+/** Max width for assembled prompt messages to avoid TUI rendering crashes on narrow terminals. */
+const MAX_PROMPT_WIDTH = 200;
 
 // NOTE: formatDenyReason, formatUserDeniedReason, and
 // formatPermissionHardStopHint have been moved to denial-messages.ts.
@@ -50,7 +54,10 @@ export function formatAskPrompt(
       fullCommand && fullCommand !== subCommand
         ? ` (full command: '${fullCommand}')`
         : "";
-    return `${subject} requested bash command '${subCommand}'${qualifierInfo}${fullCommandInfo}. Allow this command?`;
+    return truncateInlineText(
+      `${subject} requested bash command '${subCommand}'${qualifierInfo}${fullCommandInfo}. Allow this command?`,
+      MAX_PROMPT_WIDTH,
+    );
   }
 
   if (isMcpCheck(result) && result.target) {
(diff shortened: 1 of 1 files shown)


gotgenes:
Thanks for the PR, and for filing the companion fix upstream in `earendil-works/pi#7116` — that's the right place for the underlying guard.

Before taking this one, I need to ask which version you hit the crash on, because I haven't been able to reproduce it on `main`.

The inline TUI prompt has clipped every rendered line to the terminal width since e3c69070, released in `@gotgenes/pi-permission-system@20.7.1`. `PermissionPromptComponent.render()` routes all three render paths through `fitToWidth()`:

```ts
render(width: number): string[] {
  return fitToWidth(this.renderStep(), width);
}
```

`fitToWidth()` wraps with `wrapTextWithAnsi()` and then guards with `truncateToWidth()`, which is the wide-character-aware API that pi-tui's own crash message recommends. There's a regression test for it in `test/authority/permission-prompt-component.test.ts` ("clips every rendered line to the terminal width").

I also wrote a throwaway test reproducing your exact scenario against current `main` — a 279-column `ctx_batch_execute` message rendered at 213 columns, plus an 80-column terminal and a CJK message — and no line exceeded the terminal width in any case.

The one path that doesn't clip is `requestPermissionDecisionFromUi()`, but that only runs when `ctx.mode !== "tui"` (the RPC/frontend case from #519), where pi-tui isn't doing the rendering.

So my best guess is that you were on a release older than 20.7.1. Could you confirm:

1. The output of `pnpm ls @gotgenes/pi-permission-system` (or the version in your Pi extension config).
2. Whether it still reproduces on the current release.
3. If it does, the `pi-crash.log` that pi-tui writes on the throw — it records the terminal width and every rendered line's width, which would tell us exactly which line escaped.

One design note in case it does still reproduce and we need a fix here: capping the assembled message would truncate content that the component currently wraps losslessly, so the user would approve a command whose tail they can't see. That works against what #648 is asking for. A fix at this layer would want to preserve the full text and let the component keep wrapping it.

Happy to reopen the discussion as soon as I know the version.

gotgenes:
Thanks for this, and I apologize that it sat as long as it did.

The crash you reported is real and it is addressed — but through a different mechanism than this PR, so I am not going to merge it.

Your PR arrived alongside several others pulling the prompt in the opposite direction: #716 wanted the assembled message expanded into aligned `key : value` lines, and #648 and #654 wanted more added to it. Merging any one of them first would have silently set the premise the others were reviewed against. So I stopped and wrote the contract down instead, as [ADR 0011](https://github.com/gotgenes/pi-packages/blob/main/packages/pi-permission-system/docs/decisions/0011-prompt-presentation-contract.md). This PR is in its context table by number.

What §5 decided: a render is bounded by a **height budget in rows** plus a **per-field width cap**, rather than by a cap on the assembled message. Rows are the unit because the reported failure is a viewport takeover, and the per-field cap exists for exactly the case you found — one pathological field consuming the whole row budget through wrapping. The same section soft-deprecates `toolInputPreviewMaxLength`, the knob this PR extends.

So the bound now lives in the renderer rather than in `formatAskPrompt`, which is why the diff no longer applies. Your report is what made the width half of that decision concrete, and the ADR's context table is the record of it. Thank you for finding it, and for the companion fix upstream.

PR #661, opened 2026-07-26
fix(pi-subagents): hold completion nudges while the parent agent streams

daoguademeng:
## Problem

A background agent's completion nudge can be delivered **twice** — once as the pulled result mid-turn, and again verbatim at turn end.

`pi.sendMessage()` is fire-and-forget: once the nudge is handed to Pi's `followUp` queue it cannot be recalled, yet it is only *delivered* when the parent's agent loop ends. The existing guards check `record.consumed` at schedule time and at fire time — but "fire" means *handoff to Pi's queue*, not *delivery to the LLM*. Between handoff and delivery lies the entire remainder of the parent's current turn.

Reproduction timeline (100% deterministic):

| Time | Event |
|---|---|
| T | Background agent completes while the parent turn is streaming |
| T+200ms | Nudge fires → handed to Pi's followUp queue (`consumed` correctly false at that instant) |
| T+n (same turn) | Parent calls `get_subagent_result(id)` → full result returned, `markConsumed()` — too late, the message is already queued |
| Turn end | Pi delivers the queued `<task-notification>` with the same result preview; `triggerTurn` forces a redundant LLM turn re-reporting the same result |

The 200ms hold (CHANGELOG: *"pi.sendMessage() is fire-and-forget, so completion notifications … could not be retracted"*) only covers pulls within 200ms of completion — it shrinks the completion→pull window, but the dangerous window is fire→turn-end.

## Fix

`NotificationManager` now tracks the parent agent loop via `agent_start`/`agent_end`. While the parent streams, fired nudges are **held inside the extension** — where `record.consumed` can still be consulted — and flushed on agent end with a fresh consumed re-check.

Delivery timing is unchanged: `followUp` semantics already deliver at agent-loop end; the queue simply lives where it can still be cancelled. When the parent is idle, nudges send immediately at fire time exactly as before.

## Tests

- `holds a nudge that fires while the parent agent is streaming and delivers it on agent end`
- `suppresses a held nudge when the record is consumed before agent end (fire→pull→turn-end race)` ← the bug repro; fails on main
- `delivers at fire time when the parent went idle before the hold elapsed`
- `does not duplicate a held nudge when the agent re-completes before agent end`
- `dispose clears held nudges`

All 1088 tests pass; `tsc --noEmit` and `biome check` clean.

### DIFF
--- a/packages/pi-subagents/src/index.ts
+++ b/packages/pi-subagents/src/index.ts
@@ -63,6 +63,13 @@ export default function (pi: ExtensionAPI) {
     (msg, opts) => pi.sendMessage(msg, opts),
   );
 
+  // Track the parent agent loop so completion nudges are held while it streams
+  // and flushed (with a fresh consumed re-check) when it ends — a nudge handed
+  // to Pi's followUp queue mid-turn could not be recalled if the parent pulled
+  // the result later in the same turn, duplicating the delivery.
+  pi.on("agent_start", () => notifications.onParentAgentStart());
+  pi.on("agent_end", () => notifications.onParentAgentEnd());
+
   // Settings: owns all three in-memory values and handles load/save/emit.
   // onMaxConcurrentChanged is wired to the limiter directly (closure captures by reference).
   const settings = new SettingsManager({
--- a/packages/pi-subagents/src/observation/notification.ts
+++ b/packages/pi-subagents/src/observation/notification.ts
@@ -132,6 +132,14 @@ const NUDGE_HOLD_MS = 200;
 
 export class NotificationManager implements NotificationSystem {
   private pendingNudges = new Map<string, ReturnType<typeof setTimeout>>();
+  // pi.sendMessage is fire-and-forget: once a followUp is handed to Pi's queue
+  // it cannot be recalled, yet it is only delivered when the parent's agent
+  // loop ends. A parent that pulls the result (get_subagent_result) between
+  // handoff and delivery would then receive the same result twice. So while
+  // the parent is streaming, fired nudges are held here — where
+  // record.consumed can still be consulted — and flushed on agent end.
+  private heldNudges = new Map<string, Subagent>();
+  private parentStreaming = false;
 
   constructor(
     private sendMessage: (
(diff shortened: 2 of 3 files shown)


gotgenes:
Thank you, @daoguademeng, for the sharp diagnosis here — the fire→turn-end delivery window (queued via `pi.sendMessage`'s unrecallable `followUp` path, only drained at turn end) was correctly identified and reproducible on `main`.

We evaluated this PR and adopted the capability with a simplified design rather than merging the diff directly: gating nudge delivery on the `agent_settled` lifecycle event (once per agent run) instead of `agent_end` (which fires once per run *segment*, across retries/compaction), and collapsing the hold mechanism to a single pending-nudge map instead of layering it alongside the existing 200ms debounce.

Implemented in 8f7f387dc38d181589acaa5016e0a8810cbac825.

- `NotificationManager` now withholds a completion nudge that arrives while the parent's agent run is active, and flushes it on `agent_settled` with a fresh `consumed` re-check — so a result the parent already pulled mid-run is dropped instead of announced a second time.
- The 200ms `NUDGE_HOLD_MS` debounce is removed; the turn-boundary hold subsumes it (every `markConsumed()` call site already runs inside a tool handler, i.e. only while streaming).

User-visible effect: a background agent's completion no longer produces a duplicate `<task-notification>` and a redundant forced turn when the parent pulls the result via `get_subagent_result` in the same turn it completed.

This is a **breaking change** for hosts on Pi < 0.80.5: `agent_settled` doesn't exist before Pi 0.80.4 (never published to npm — the registry jumps 0.80.3 → 0.80.5), so `@gotgenes/pi-subagents` now requires `@earendil-works/pi-coding-agent >= 0.80.5`, raised from `>= 0.75.0`. This ships as `fix(pi-subagents)!:` with a `BREAKING CHANGE:` footer.

Closing this PR in favor of the independent re-implementation, per the review triage — full reasoning in `packages/pi-subagents/docs/retro/0661-hold-completion-nudges-while-parent-streams.md`. Every implementation commit carries your `Co-authored-by` trailer. Thanks again for tracking this down!

PR #662, opened 2026-07-26
fix(pi-subagents): honor wait:true for queued agents and in-flight resumes

daoguademeng:
## Problem

`get_subagent_result`'s wait path only awaits `status === "running"`, silently returning a stale report in two cases the codebase already promises to support:

**1. Queued agents.** `scheduleVia` captures the limiter promise eagerly — its own doc comment says *"a still-queued agent is awaitable via the `promise` getter from spawn"* — but the tool never uses it. `wait: true` on a queued agent returns immediately with a useless "queued" report instead of waiting for the slot to free and the run to finish.

**2. In-flight resumes.** `Subagent.resume()` never refreshed `_promise`, so the `promise` getter kept the settled handle of the *original* run. `wait: true` during a background resume awaited an already-resolved promise and reported "running".

## Fix

- The wait guard now uses `record.isActive()` (running **or** queued) — one-line change plus a comment documenting why queued is safe.
- `resume()` stores the live resume promise in `_promise` before returning (extracted `runResume` keeps the always-resolves contract), so waiters track the current run. Synchronous-throw behavior for a missing session is unchanged.

## Tests

- `waits for a queued agent when wait=true (promise is captured at spawn)` — simulates limiter admission after the wait begins; fails on main
- `exposes the in-flight resume via the promise getter (wait support)` — asserts the promise doesn't settle before the resume does; fails on main

All 1085 tests pass; `tsc --noEmit` and `biome check` clean.

### DIFF
--- a/packages/pi-subagents/src/lifecycle/subagent.ts
+++ b/packages/pi-subagents/src/lifecycle/subagent.ts
@@ -343,6 +343,14 @@ export class Subagent {
 			throw new Error("Subagent not configured for resume — missing session");
 		}
 
+		// Refresh the awaitable handle: waiters (e.g. get_subagent_result wait:true)
+		// must track the live resume, not the settled promise of a prior run.
+		const run = this.runResume(subagentSession, prompt, signal);
+		this._promise = run;
+		return run;
+	}
+
+	private async runResume(subagentSession: SubagentSession, prompt: string, signal?: AbortSignal): Promise<void> {
 		this.resetForResume(Date.now());
 		this.listeners.attachObserver(subscribeSubagentObserver(subagentSession, this.state, {
 			onCompact: (info) => this.execution.observer?.onCompacted?.(this, info),
--- a/packages/pi-subagents/src/tools/get-result-tool.ts
+++ b/packages/pi-subagents/src/tools/get-result-tool.ts
@@ -32,8 +32,10 @@ export class GetResultTool {
 			return textResult(`Agent not found: "${params.agent_id}". Records are cleared at session start/switch, so it may be from a previous session.`);
 		}
 
-		// Wait for completion if requested.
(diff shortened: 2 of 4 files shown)


gotgenes:
Thanks for the report and the repro, @daoguademeng — both defects you flagged were real (confirmed by reverting just the `src/` changes against current `main` and watching the two new tests fail).

We took the direction rather than the diff: same two fixes, plus closing a gap the review surfaced (`GetResultTool.execute` was ignoring its own `AbortSignal`, which matters once a `wait: true` call can block for the whole queue depth, not just one agent's run).

Implemented in 908dadccf188234faa1f886cd8ade17ed134f267 (`main`), across:

- fb298c6ce66bff9e974a59f8d5c070f7c245c5ea — `fix(pi-subagents): honor wait:true for queued agents`
- 97abfa1d5467e18fa7e14ebb0d7a4f784b114e8c — `fix(pi-subagents): end a get_subagent_result wait on parent interrupt`
- ceb8234ef4205ddd6c9d9cee46c5875156bcadf1 — `fix(pi-subagents): track the live resume in the Subagent promise getter`

User-visible change: `get_subagent_result` with `wait: true` now actually waits for a queued agent (not just a running one) and for an in-flight `resume()`, and a parent-turn interrupt ends the wait without cancelling the agent. Not breaking — no default, config key, or output shape changed.

Every implementation and docs commit carries a `Co-authored-by: daoguademeng <whumaple@gmail.com>` trailer.

Closing this in favor of the commits above rather than merging, per our third-party PR review process (triage note: `packages/pi-subagents/docs/retro/0662-honor-wait-for-queued-and-resuming-agents.md`). Appreciate the report — feel free to open a fresh issue/PR for anything else you're seeing.

Issue #639, opened 2026-07-23
pi-permission-system: decide the permission policy model — capabilities, config shape, prior art (ADR 0013)


gotgenes:
### Package

pi-permission-system

### What do you want to change?

Author ADR 0009 deciding the future shape of the permission policy model: whether and how access capabilities (reading vs. writing paths, exec, net) become first-class in policy, what the config keys should look like (flat keys, nesting, naming), and what composition semantics we want.

This came out of planning #609. That issue asks for output redirects to be governed separately from the command; exploring it showed the gap is general — access direction/capability is not a first-class fact anywhere in the model. Rather than bolt on one key, I want a deliberate decision about the model itself.

Nothing is locked down going in, including the current config format. The decision criteria: clarity, simplicity (straightforward, avoiding a complex calculus of interactions between rules), and no ambiguity — for our users first.

### Why?

The fork's history is a long catalog of capability escaping pattern matching through syntactic synonyms: whole-string bypasses (#301), substitutions (#306), `cd` folding (#393), wrapper opacity (#481, #490, #575), env prefixes, unparseable commands failing closed (#452), and now redirects (#609). Command-pattern rules cannot close that class; each fix moves the frontier.

We are never going to beat an agent sandbox at containment. What we can be is the best at routing human attention to consequential actions — and consequence is about effects (read vs. write vs. exec vs. net), not command spellings. OpenCode's model, which we started from, is actor-keyed and cannot see this distinction; ours currently cannot either.

### How? (optional)

A deliberative ADR, settled interactively (the #581 lesson — the deliberation is the deliverable, not transcription of prior sketches):

- Survey prior art with citations: OpenCode, Deno permissions, Node's permission model (`--allow-fs-read`/`--allow-fs-write`), Linux Landlock, macOS Seatbelt, WASI preopens, systemd sandboxing — what each chose for naming, composition semantics, defaults, and unknown-handling.
- State the threat model explicitly: a cooperative-but-fallible agent; attention routing, not containment; a future sandbox as the enforcement seam.
- Evaluate the option space: status quo plus a targeted `path_write` key; a capability family of flat keys; nested facets under `path`; an effect-centered model with a command-effects knowledge base; a full config remodel.
- Decide staging: what #609's implementation builds first.

Issue #609 stays open and gets re-planned once the ADR lands.

gotgenes:
[...]

PR #671, opened 2026-07-27
feat(pi-permission-system): add launcher yolo override

gavinsomers:
## Summary

- add the exact `PI_PERMISSION_SYSTEM_YOLO=1` launcher contract
- apply it as an in-memory runtime override across startup and config refresh
- keep explicit deny rules enforced and prevent the override from being persisted
- document the process-scoped compatibility boundary

## Verification

- `corepack pnpm --filter @gotgenes/pi-permission-system test` (131 files, 2685 tests)
- `corepack pnpm --filter @gotgenes/pi-permission-system run check`
- `corepack pnpm run lint`
- `corepack pnpm fallow dead-code`

## Integration

This is the package side of the FirstMate `+yolo` integration tracked as OPS-232. FirstMate will set the variable only for delegated Pi ship/scout processes in projects whose effective mode is `+yolo`; yolo-off, secondmate, non-Pi, and standalone Pi behavior remains unchanged.

### DIFF
--- a/packages/pi-permission-system/README.md
+++ b/packages/pi-permission-system/README.md
@@ -118,6 +118,11 @@ The optional `authorizerChain` field names registered case-by-case decision link
 A downstream extension registers a link via `getPermissionsService().registerAuthorizer(name, authorize)`; it decides nothing until you name it here (opt-in), config order fixes the chain order, and the chain owner caps any link's `allow` on `external_directory`/`path` to keep it within your policy — see [docs/configuration.md](docs/configuration.md#authorizer-chain--case-by-case-decision-links).
 [`@gotgenes/pi-permission-model-judge`](https://github.com/gotgenes/pi-packages/tree/main/packages/pi-permission-model-judge) is a first-party reference implementation of such a link — a deny-first reviewer that auto-denies mistyped out-of-directory paths.
 
+Trusted launchers can enable the existing deny-preserving yolo composition for one Pi process with `PI_PERMISSION_SYSTEM_YOLO=1 pi`.
+Only the exact value `1` enables the override, configuration reloads retain it, and it never changes global or project config files.
+The override converts `ask` results to `allow` while explicit `deny` rules remain enforced.
+See [the launcher-scoped yolo reference](docs/configuration.md#launcher-scoped-yolo-override) for the complete contract.
+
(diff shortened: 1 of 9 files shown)


gavinsomers:
Companion FirstMate integration: [kunchenguid/firstmate#1129](https://github.com/kunchenguid/firstmate/pull/1129).

That PR scopes `PI_PERMISSION_SYSTEM_YOLO=1` to delegated Pi ship/scout launches for effective `+yolo` projects and pins this package PR at commit `79829ffd`.

gotgenes:
Thanks for this, and for the thorough test coverage — I've approved the workflow run and CI is green.

I don't want to leave you waiting without a status, so: I'm not going to be able to give you a yes or no on this quickly. Introducing an environment variable that relaxes the permission policy is a design decision rather than a code review, and it's one I want to make deliberately rather than in a single PR thread.

That decision is tracked in #639, where I'm working out the permission policy model — what's configurable, through which channels, and with what precedence. Several open requests (#603, #604, #680, and #684) are all asking for related widenings, and I'd rather settle the model once than answer each one ad hoc and end up with an inconsistent surface.

So this is parked pending #639, not declined and not forgotten. I'll come back to this PR with a real answer once that lands, and I'll make sure the reasoning is written down rather than just asserted.

One thing that would help in the meantime: could you say a bit more about the FirstMate integration's constraints? Specifically, whether the launcher is able to write to the operator's global config, or whether the process environment is genuinely the only channel available to it. That distinction matters for which shapes are on the table.

gotgenes:
A status update on the park, since two things have changed since my last comment.

**#639 landed; the half your PR waits on didn't.** #639 closed as [ADR 0013](https://github.com/gotgenes/pi-packages/blob/main/packages/pi-permission-system/docs/decisions/0013-permission-policy-model.md), which settled the capability model and config shape. It deliberately didn't settle *which channels* policy may enter through — that split off into #799, and an environment variable relaxing the policy is squarely that question. So this is still parked, now on #799 rather than #639.

**The tree moved underneath your branch.** #837 reorganized `pi-permission-system/src`, moving the 42% of files that sat at the package root into subdirectories. Two of the three `src` files this PR touches were moved, so rebasing isn't worth your time right now — I'd leave the branch as it is until #799 is decided.

Thanks again for the FirstMate context on the companion PR; it's on the pile of evidence #799 will weigh.

PR #675, opened 2026-07-27
feat(pi-permission-system): load policy from Pi settings

chronoAP:
## Summary
- load global and project permission policy from Pi settings.json under piPermissionSystem
- support per-agent policy through piPermissionSystem.agents.<name>
- preserve dedicated config and frontmatter precedence, project-trust gating, strict validation, and cache invalidation
- document the settings schema and complete precedence order

## Validation
- pnpm --filter @gotgenes/pi-permission-system check
- 55 focused tests passed across settings loading, policy integration, policy loader, config paths, and config reporting
- targeted Biome, ESLint, and rumdl checks passed
- independent reviewer found no actionable issues

## Environment note
The full suite was also attempted on Windows; it is not clean in this checkout due to 112 platform/path and POSIX file-mode failures. The focused tests for this change and TypeScript checks pass.


gotgenes:

**It's parked, and on what.** Loading policy from Pi settings is precisely the question of which channels permission policy may enter through, and with what precedence and trust gating. That's being settled in #799 rather than in any one PR thread. Several open requests — #603, #604, #691, #720, and PRs #638, #671, #684, #692 — are all asking for related widenings, and I'd rather answer them once from a written decision than one at a time and end up with an inconsistent surface.

**The tree moved underneath your branch.** #837 reorganized `pi-permission-system/src`, moving the 42% of files that sat at the package root into subdirectories. Nine of the ten `src` files this PR touches were moved, so rebasing isn't worth your time right now — once #799 is decided this will mean reapplying the change against the new layout rather than resolving conflicts.

Thanks for your patience.

PR #681, opened 2026-07-28
fix(pi-subagents): replace turn glyph (#669)

### LINKED ISSUE #669: pi-subagents: formatTurns() glyph U+27F3 (⟳) overflows its monospace cell, overlapping the next column
### Package

pi-subagents

### Package version

19.0.0

### What happened?

This is the same issue as https://github.com/tintinweb/pi-subagents/issues/84.

Below is an example of a subagent rendering sub-optimally, due to the use of U+27F3 in formatTurns().
I believe that either switching to ASCII, or the same unicode character in the fix would be sufficient.

Image

### Expected behavior

The turn glyph in the formatTurns() should not be overflowing the terminal column.

### Steps to reproduce

TBH I'm not sure what exactly makes this reproducible on my machine, yet not on others.
I use Ghostty as my terminal emulator, with IBM Plex Mono and IBM Plex Sans KR as my fallback font, and also my macOS native locale is set to `ko_KR.UTF-8` (though I do `export LANG=en_US.UTF-8` in all of my terminals) Maybe it's that the CJK fallback font is rendering it?

### Additional context

See above.

goranmoomin:
Fixes the pi-subagents turn-count glyph overflow in monospace terminals.

This is directly from the upstream fix for tintinweb/pi-subagents#84 (https://github.com/tintinweb/pi-subagents/commit/aad202a3583c8008d03760be8e8ac4323d38bec7).

- Replace the turn-count glyph from `⟳` to `↻`.
- Move the compaction-count glyph from `↻` to `⇊` to keep the indicators distinct.
- Update rendered-output tests and README examples.

This is a visual-only output change.

Fixes #669.

### DIFF
--- a/packages/pi-subagents/README.md
+++ b/packages/pi-subagents/README.md
@@ -65,11 +65,11 @@ The extension renders a persistent widget above the editor showing active backgr
 
 ```text
 ● Agents
-├─ ⠹ Agent  Refactor auth module · ⟳5≤30 · 5 tool uses · 33.8k token (62%) · 12.3s
+├─ ⠹ Agent  Refactor auth module · ↻5≤30 · 5 tool uses · 33.8k token (62%) · 12.3s
 │    ⎿  editing 2 files…
-├─ ⠹ Explore  Find auth files · ⟳3 · 3 tool uses · 12.4k token (8%) · 4.1s
+├─ ⠹ Explore  Find auth files · ↻3 · 3 tool uses · 12.4k token (8%) · 4.1s
 │    ⎿  searching…
-├─ ⠹ Agent  Long-running task · ⟳42 · 38 tool uses · 91.0k token (84% · ↻2) · 2m17s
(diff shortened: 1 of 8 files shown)


gotgenes:
Thanks for this, @goranmoomin — and for tracking down that it's the upstream fix for tintinweb/pi-subagents#84.

I re-implemented rather than merging this directly: I wanted to centralize the package's scattered glyph literals (some written as literal characters, some as `\uXXXX` escapes, sometimes both in the same file) into one `src/ui/glyphs.ts` module before making the substitution, so a future glyph change can't miss a site the way a plain grep would. Landed the same glyph choices you and upstream settled on — `↻` for turns, `⇊` for compactions — in 8a54b9cabac1d2535e865a9a46ddb8f6845e21aa, on `main` now. Full writeup in the closing comment on #669.

I also filed #683 to track the same defect class for `⎿`, the Braille spinner, and `◍`, which have similar-to-worse monospace coverage but are Pi/Claude-Code house style rather than this package's own choice.

Closing this in favor of the above, with credit for both the fix and the root-cause analysis. Thank you again!

Issue #620, opened 2026-07-19
pi-permission-system: allow-capable opaque-bash adjudicator (Authorizer slice 2)


gotgenes:
## What

Ship the **allow-capable** side of the model-judge Authorizer chain — ADR 0007's slice 2 (the opaque-bash adjudicator), gated behind the full bounded-delegation envelope.

Continues #472 (Track B); follows #599 (Step 5, which lands the `registerAuthorizer` seam, the `authorizerChain` config, the injected `PermissionQuery`, and a **conservative** enforcement checkpoint).

## Why

#599 ships the chain seam with two deliberately conservative properties, because its only day-one consumer (#600) is deny-first and never allows:

- The enforcement checkpoint excludes the **whole** `path` surface (any link `allow` on `path`/`external_directory` downgrades to `defer`), rather than only secret-shaped paths. There is no formal secrets model in the codebase, so a hard-coded secret denylist was declined as a bypass surface shipped ahead of any consumer.
- The injected `PermissionQuery` (`authorize(details, query)`) has no consumer yet — the deny-first reviewer decides by pattern, never by querying the engine.

This issue is where the allow-capable capability — and therefore both of those refinements — actually earns its place.

## Proposed change

Per ADR 0007 §6 (the capability gradient) and §3 (tool-augmented decomposition):

- An allow-capable adjudicator link that decomposes an opaque bash command (`bash -c "…"`, `eval`, unparseable constructs the parser floors to `ask` via sentinels) and queries the deterministic engine per sub-command through the injected `PermissionQuery`, allowing only what the engine already grants for the pieces it identifies.
- Refine the enforcement checkpoint: relax the whole-`path` exclusion introduced in #599 to a **secret-shaped** exclusion, ideally driven by the operator's own `path`/`external_directory` deny/ask rules rather than a hard-coded denylist, so a link may allow a non-secret path but never a secret-shaped one.
- Audit provenance: grants from a model link record `origin: "authorizer:model"` and are **non-persistent** (never `approved_for_session`).
- The full envelope per ADR 0007 §6: hard exclusions, off by default, bounded-delegation ruleset, decomposition-fidelity as the residual risk.

## Context

- Design: `docs/decisions/0007-model-judge-authorizer-chain-adr.md` (§3, §6), accepted.
- Depends on #599 (the seam, config, injected query, conservative checkpoint).
- Part of #472; the deny-first slice is #600.
[...]

PR #684, opened 2026-07-28
feat(pi-permission-system): allow authorizers to approve external_directory access (opt-in)

alexander-potemkin:
## Summary

Adds an opt-in config flag **`allowAuthorizerOnExternalDirectory`** (default `false`) to the bounded-delegation checkpoint (ADR 0007 §5).

When `true`, the checkpoint removes `external_directory` from its excluded-surface set, so a named chain link (e.g. an LLM auto-review judge registered via `registerAuthorizer` and named in `authorizerChain`) may grant `allow` on outside-CWD asks instead of always downgrading the verdict to `defer` and falling through to the terminal prompt. The `path` surface stays excluded **regardless**, so secret-shaped paths (`.env`, `~/.ssh/*`) can never be auto-approved by a link.

## Why

The current `DELEGATION_EXCLUDED_SURFACES = { external_directory, path }` makes it impossible for a model judge to ever auto-approve outside-working-directory access — every outside-CWD ask always prompts the user. The maintainer already flags this as conservative in the source (deferred to #620). Making the `external_directory` exclusion opt-in enables a Codex/Claude Code-style auto-review flow in Pi, where a judge decides outside-CWD access while secrets remain hard-protected by the retained `path` exclusion.

The flag defaults to `false`, so existing behavior is unchanged; it only relaxes the whole-`external_directory` exclusion, and operators are expected to pair it with a scoped `external_directory` permission policy.

## Changes

- `delegation-envelope.ts`: `encloseInDelegationEnvelope(authorize, allowAuthorizerOnExternalDirectory = false)`; keeps `path` excluded, drops `external_directory` from the excluded set when enabled.
- `authorizer-selection.ts` / `index.ts`: thread the config value (`configStore.current().allowAuthorizerOnExternalDirectory === true`) into the envelope.
- `extension-config.ts` / `config-schema.ts` / `config-loader.ts` / `schemas/permissions.schema.json`: new boolean field with `default: false`, merged as a boolean scalar.
- `docs/architecture/architecture.md`: note the opt-in relaxation.

## Test plan

- `delegation-envelope.test.ts`: an `allow` on `external_directory` passes through when the flag is `true`; an `allow` on `path` is still capped to `defer`; the cap is retained when the flag is `false`.
- `authorizer-selection.test.ts` / `forwarded-request-server.test.ts`: updated for the new deps field and existing single-arg call sites (default `false`).
- `pnpm --filter @gotgenes/pi-permission-system run check`, `run lint`, and `run test` all pass (2675 tests).

🤖 Generated with [OpenCode](https://opencode.ai)


gotgenes:

**It's parked, and where the question is being decided.** Letting an authorizer approve `external_directory` access changes what a chain link is permitted to decide, which [ADR 0007](https://github.com/gotgenes/pi-packages/blob/main/packages/pi-permission-system/docs/decisions/0007-model-judge-authorizer-chain-adr.md) §5 currently excludes. I've opened #882 to decide that specific question rather than answer it in a PR thread, and it's scheduled as a step in the package's current improvement phase. The broader question of which channels may widen policy at all is #799. I can't tell you yet which way #882 will go, only that it'll be answered in writing.

**The tree moved underneath your branch.** #837 reorganized `pi-permission-system/src`, moving the 42% of files that sat at the package root into subdirectories. Four of the six `src` files this PR touches were moved, so rebasing isn't worth your time right now — I'd leave the branch as it is until #882 is decided.

Thanks for your patience.

alexander-potemkin:
Thank you and never mind!
To be honest, I've ended up actually spending most of my time in omp, instead of pi directly.

PR #690, opened 2026-08-01
fix(pi-subagents): make the /subagents:sessions preview responsive on large transcripts

### LINKED ISSUE #689: pi-subagents: /subagents:sessions becomes unresponsive on large live transcripts
## Observed behavior

On `@gotgenes/pi-subagents` 19.2.1, opening `/subagents:sessions` for a long-running live agent makes the overlay progressively unresponsive:

- Representative session: **165 tool uses / 267.5k tokens / 10,876 rendered transcript lines**.
- Rich assistant output streams far more slowly than the parent Pi TUI.
- Up/Down and PgUp/PgDn appear frozen. Input is accepted, but repaint blocks long enough to look dead.
- Pressing Esc immediately restores the fast one-line Agents widget and normal TUI responsiveness.

## Root cause

Two O(total transcript) paths stack inside `TranscriptOverlay`:

1. `liveSource.subscribe` discards `AgentSessionEvent`, and the overlay calls `rebuild()` for every event—including each `message_update` text/thinking/tool-call delta and `tool_execution_update`. This recreates every Pi rich component in the full transcript per chunk.
2. Both `render()` and `handleInput()` call `buildContentLines()`, which invokes `Container.render()` and truncates/maps all transcript lines before slicing the viewport. Every paint and keypress therefore rerenders 10k+ rows.

There is also a width mismatch: `handleInput()` lays out at 100% terminal width (`tui.terminal.columns - 4`), while the overlay actually renders at 90%. A naïve one-width cache would therefore perform two full renders per keypress—100% during input, then 90% during paint—and scroll bounds are calculated against the wrong layout.

Pi's main TUI does not rebuild history per token; it updates only the current `AssistantMessageComponent`.

## Reproduction and measurements

A deterministic Vitest fixture with 200 multiline messages (approximately 4k content rows) shows:

- Current `main`: initial paint = 201 `Container.render()` calls; same-width paint + Up + paint = **804** total, meaning three complete rerenders after the initial one.
- Fixed path: the count remains **201** while scroll output still changes, proving the key works without another rich-history render.
- Old-style rebuild + paint: 20 cycles = **263.6 ms** (approximately 13.2 ms/event).
- Fixed hot path: 100 scroll + paint cycles = **29.1 ms** (approximately 0.29 ms/interaction), about **45× lower per operation** on the synthetic fixture.

The reported real session is both larger and tool-component-heavy, so its gap is expected to be worse.

## Proposed fix

A tested implementation is available on [`daoguademeng/pi-packages@fix/session-preview-performance`](https://github.com/daoguademeng/pi-packages/tree/fix/session-preview-performance), commit [`a4a8caf`](https://github.com/daoguademeng/pi-packages/commit/a4a8caf):

- Forward `AgentSessionEvent` through `TranscriptSource`.
- Cache settled transcript lines by the compositor's actual overlay width.
- Slice directly across settled/live rows, making paint and scrolling O(viewport).
- Keep the current assistant message as a separate rich `AssistantMessageComponent`, updated per delta like Pi's main TUI. This preserves full live thinking/text rather than freezing it until `message_end`.
- Rebuild settled history only when a message settles, the agent ends, or compaction replaces history.
- Invalidate both width caches correctly.

## Verification

- 64 test files / **1,141 tests passed**.
- `tsc --noEmit` passed.
- Full package Biome, ESLint, and Markdown lint passed.
- A real `pi -ne -e ...` end-to-end subagent spawn returned `PREVIEWOK` using `openai-codex/gpt-5.4-mini`.

daoguademeng:
Fixes #689.

## What this changes

`/subagents:sessions` became progressively unresponsive on large live transcripts (reference session: 165 tool uses / 10,876 rendered lines): rich output streamed far slower than the parent TUI, and arrow keys appeared frozen. Two commits, reviewable independently:

### 1. `fix: make session preview rendering viewport-bound`

- Forward `AgentSessionEvent` through `TranscriptSource.subscribe` instead of discarding it.
- Cache settled transcript lines keyed by the compositor's **actual overlay width**; paint and scroll slice the cache instead of re-rendering and re-truncating every line per keypress.
- Keep the in-progress assistant message as its own `AssistantMessageComponent`, updated per delta — the same model as Pi's interactive mode. Full live thinking/text fidelity is preserved; deltas never touch settled history.
- `handleInput` reuses the width the overlay actually rendered at, fixing the 100%-vs-90% width mismatch (scroll bounds were computed against a layout that was never displayed). This is the same root cause independently diagnosed in #670 by @aqua2k1 — this commit covers it as part of the broader change; happy to rebase if #670 lands first.

### 2. `perf: settle session-preview messages incrementally`

After commit 1, every settling event (`message_end` fires for each assistant message **and each tool result**) still rebuilt the full component tree and re-rendered all lines: ~101 ms per settle on a 200-message fixture at production width, i.e. a visible stutter per tool completion on large sessions.

- Hold the settled transcript as per-message blocks, each caching its rendered lines. New messages append blocks; a tool result mutates only its paired tool component's block; the flat line array is a cheap concat.
- Correctness rests on a verified runtime invariant: agent-core reduces state before awaiting listeners (`Agent.processEvents` pushes into `state.messages` before `message_end` listeners run), so consuming at event time always sees the settled prefix; the streaming message lives outside `state.messages`, so live and settled never overlap.
- An identity guard on the last consumed message falls back to a full rebuild whenever history is rewritten wholesale (compaction, branching, or sources that fabricate message arrays per call). `agent_end` and `compaction_end` keep full-rebuild semantics to capture in-place mutations (e.g. abort `errorMessage` patches).
- Known boundary, unchanged from before: `custom`-role messages are still skipped (rendering them needs the child session's message-renderer registry).
- The implementation stays below the repository's Fallow complexity gates without suppressions: scroll input is table-driven, message consumption dispatches by role, and source-event routing uses narrow helpers.

## Measurements

Deterministic Vitest fixture, 200 multiline messages (~4.4k rendered lines), production width 144:

| Hot path                       | before                 | after commit 1 | after commit 2     |
| ------------------------------ | ---------------------- | -------------- | ------------------ |
| Settle (`message_end` + paint) | ~101 ms (full rebuild) | ~101 ms        | **1.06 ms (~95×)** |
| Streaming delta + paint        | full rebuild per delta | 1.56 ms        | **0.43 ms**        |
| Scroll + paint                 | full re-render per key | 0.09 ms        | **0.09 ms**        |

The reported 10.9k-line session is ~2.5× this fixture and tool-component-heavy, so absolute gains there are larger.

## Verification

- TDD, red → green on both commits. New tests pin: O(changed-message) settle render counts, tool-result updates touching only the affected block, **line-for-line equivalence between incremental settles and a freshly built overlay**, `agent_end` capturing in-place mutations, the rewritten-history fallback, 90%-overlay-width input behavior, width-change invalidation, and full-fidelity live thinking/text streaming.
- Suite: 64 files / 1,146 tests green; `tsc --noEmit`, Biome, ESLint, and markdown lint clean.
- `pnpm fallow audit --base origin/main` clean with no complexity suppressions; GitHub Actions `check` passes.
- End-to-end: real `pi -ne -e src/index.ts` sessions spawning subagents through this branch returned expected results (foreground and background).

As with the previous rounds — take the direction and reimplement if that fits your process better; the diagnosis and tests should transfer either way.

### DIFF
--- a/packages/pi-subagents/src/ui/session-navigation.ts
+++ b/packages/pi-subagents/src/ui/session-navigation.ts
@@ -71,7 +71,7 @@ export interface TranscriptSource {
   /** Current message history. */
   getMessages(): readonly SessionMessage[];
   /** Subscribe to changes; returns an unsubscribe, or undefined for a static snapshot. */
-  subscribe(onChange: () => void): (() => void) | undefined;
+  subscribe(onChange: (event?: AgentSessionEvent) => void): (() => void) | undefined;
   /** Running-agent streaming state, or undefined when not streaming. */
   streaming(): StreamingState | undefined;
   /** Resolve a registered tool definition by name, for Pi's tool-execution components. */
@@ -129,7 +129,7 @@ export function fileSnapshotSource(
 export function liveSource(record: NavigableSubagent): TranscriptSource {
   return {
     getMessages: () => record.agentMessages,
-    subscribe: (onChange) => record.subscribeToUpdates(() => onChange()),
+    subscribe: (onChange) => record.subscribeToUpdates(onChange),
     streaming: () =>
       isRunningStatus(record.status)
         ? { activeTools: record.activeTools, responseText: record.responseText }
--- a/packages/pi-subagents/src/ui/session-navigator.ts
+++ b/packages/pi-subagents/src/ui/session-navigator.ts
@@ -15,6 +15,7 @@
  * snapshot (`fileSnapshotSource`) swaps in without touching the renderer or the overlay.
  */
 
+import type { AssistantMessage } from "@earendil-works/pi-ai";
 import {
   AssistantMessageComponent,
   BashExecutionComponent,
@@ -38,7 +39,7 @@ import {
   visibleWidth,
 } from "@earendil-works/pi-tui";
 import type { AgentConfigLookup } from "#src/config/agent-types";
-import type { SessionMessage } from "#src/types";
+import type { AgentSessionEvent, SessionMessage } from "#src/types";
 import { describeActivity, type Theme } from "#src/ui/display";
 import { GLYPHS } from "#src/ui/glyphs";
 import { fileSnapshotSource, listNavigableAgents, liveSource, type NavigableSubagent, type TranscriptSource } from "#src/ui/session-navigation";
@@ -130,11 +131,11 @@ export class SessionNavigatorHandler {
 /**
  * Read-only scrollable transcript overlay.
  *
- * Caches a `Container` of Pi's per-entry components and rebuilds it only when the
- * source changes (live agents) — each paint reuses the cached tree, so markdown
- * highlighting does not re-run per frame. This class owns scroll state, chrome,
- * and the running-agent streaming indicator; the component mapping lives in
- * `buildTranscriptComponents`.
+ * Caches the settled transcript's rendered lines per overlay width and rebuilds
+ * Pi's rich per-entry component tree only when messages settle. Streaming deltas
+ * use a lightweight activity row, so paint/scroll stay O(viewport) instead of
+ * O(total transcript). This class owns scroll state, chrome, and live activity;
+ * component mapping lives in `buildTranscriptComponents`.
  */
 export class TranscriptOverlay implements Component {
   private scrollOffset = 0;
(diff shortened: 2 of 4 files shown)


gotgenes:
Thank you, @daoguademeng — the diagnosis, reproduction fixture, and O(total) analysis here are exactly right, and the fix I landed matches your two-commit shape almost exactly (viewport-bound caching, then incremental settling).

Landed in #689 as `fix(pi-subagents): scroll the session preview at the width it renders at`, `refactor(pi-subagents): extract transcript content from the session overlay`, `perf(pi-subagents): make session-preview paint and scroll viewport-bound`, `perf(pi-subagents): update only the live message on session-preview deltas`, and `perf(pi-subagents): settle session-preview messages incrementally` — implemented in 1baf9ab99b4db5af335affdbabe290d5cb4f2762 on `main`.

Two differences from this PR, both found by tracing Pi's own agent-core source rather than a design preference: the in-flight message is never in `state.messages` (it lives in `state.streamingMessage` until it settles), so the `isLiveMessage`/timestamp-matching guard this PR carries turned out to be unnecessary; and an extension's in-place `message_end` replacement is applied by Pi before session listeners are notified, so the settled content is always final without extra handling.

This repo reimplements adopted third-party changes through its own TDD cycle rather than merging directly, so I'm closing this without merging — not a reflection on the work, which was correct and did the hard diagnostic part. Thanks again for the clear repro and the measurements, which made verifying the fix straightforward.

PR #697, opened 2026-08-02
feat(pi-subagents): exclude package extensions from children

### LINKED ISSUE #696: pi-subagents: allow children to exclude selected package extensions
## Problem

`pi-subagents` creates in-process child `AgentSession`s and reloads the parent's Pi packages for every child. Some package extensions are intentionally parent-scoped or expensive to initialize per session, but current releases provide no way to prevent those extensions from loading in children.

A concrete failure is `@cortexkit/pi-magic-context`: four concurrent children each initialize Magic Context and scan the Pi session store in the same V8 heap. With 1,396 sessions (~1.1 GB), this reproducibly reaches the ~4.1 GB V8 heap limit and aborts with exit 134. Related upstream report: cortexkit/magic-context#247.

Neither side's current settings solve this integration boundary:

- `pi-subagents` children always inherit all parent extensions.
- Magic Context's `pi.subagent_extensions` controls subprocesses created by Magic Context itself, not in-process children created by `pi-subagents`.

## Requested behavior

Add a narrow opt-out that filters package **extensions** before the child `ResourceLoader` runs, while preserving the package's skills and other resources.

For example, in `~/.pi/agent/subagents.json`:

```json
{
  "excludedExtensionPackages": [
    "npm:@cortexkit/pi-magic-context"
  ]
}
```

Package sources should use Pi's exact configured source strings. Parent behavior must remain unchanged, and an absent/empty setting must preserve today's child behavior.

## Validation target

With the exclusion enabled:

- the parent still loads Magic Context;
- children continue loading all non-excluded extensions and skills;
- four concurrent children do not initialize Magic Context or trigger the OOM path.

beilo:
## Summary

- add `excludedExtensionPackages` to layered `subagents.json` settings
- pass a child-local filtered `SettingsManager` into `ResourceLoader` before package discovery
- disable only matching package extensions; keep package skills and other resources available
- document exact Pi package-source matching and the parent/child behavior

Closes #696.

## Why

In-process child sessions currently reload every parent package extension. Extensions that are parent-scoped or expensive per session can initialize repeatedly in the shared V8 heap. A concrete reproducer is cortexkit/magic-context#247: four concurrent children load Magic Context five times total (parent + four children) and abort near the V8 heap limit while scanning a large Pi session store.

This change adds a narrow, opt-in prevent-load seam without restoring the old broad `extensions`/`isolated` agent frontmatter policy.

Example:

```json
{
  "excludedExtensionPackages": [
    "npm:@cortexkit/pi-magic-context"
  ]
}
```

## Validation

- `pnpm run check`
- `pnpm run lint`
- `pnpm test` — 64 files, 1,141 tests passed
- regression run with four concurrent children: all four returned; process remained around 60 MB RSS after five minutes instead of reproducing the previous ~4.14 GB V8 OOM / exit 134

### DIFF
--- a/packages/pi-subagents/README.md
+++ b/packages/pi-subagents/README.md
@@ -304,6 +304,25 @@ Individual projects can still override via `/subagents:settings`.
 
 **Failure behavior:** missing file is silent; malformed JSON logs a `[pi-subagents] Ignoring malformed settings at …` warning to stderr; invalid/out-of-range field values are dropped per-field; write failures downgrade the `/subagents:settings` toast to a warning with `(session only; failed to persist)`.
 
+### Excluding package extensions from children
+
+Some session-oriented extensions must run only in the parent.
+Add their exact Pi package sources to `excludedExtensionPackages`:
+
+```json
+{
+  "excludedExtensionPackages": [
+    "npm:@cortexkit/pi-magic-context"
+  ]
+}
+```
+
+The setting disables only the matched package's extensions in child resource loaders, before extension factories run.
+Other resources from that package and all resources from other packages remain available.
+The parent session is unchanged.
+
+This setting is edited directly in the global or project `subagents.json`; it is not exposed in `/subagents:settings`.
+
 ### Abort on interrupt
 
 By default, pressing ESC to interrupt the parent agent also aborts every subagent.
@@ -343,7 +362,8 @@ The earlier `isolation: "worktree"` spawn flag and `isolation:` frontmatter key 
 ## Removed: agent memory and skill preloading
 
 Persistent agent memory (the `memory:` frontmatter key) and skill preloading (the `skills:` frontmatter key) were removed when the core was slimmed down.
-Children now always inherit the parent's skills and extensions, so the `isolated`, `extensions`, and `skills` frontmatter keys no longer exist.
+Children inherit the parent's skills and extensions by default, so the `isolated`, `extensions`, and `skills` frontmatter keys no longer exist.
+Package-level extension opt-outs belong in `excludedExtensionPackages` settings instead of agent frontmatter.
 
 ## Migrating from `disallowed_tools`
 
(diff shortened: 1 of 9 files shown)


gotgenes:
Thank you for the detailed report and reproduction, @beilo — the OOM you traced through Magic Context's per-session store scan was exactly right, and the `excludedExtensionPackages` shape and exact package-source matching you proposed here carried through.

I reimplemented it with a narrower design in f847ebfebe8942ca64146fc50530d3e682819edb (building on f83e7066306ba3beaaf85f0822b0213246c8c31e and 4b167d680c79973ab61c690c6c784c30241cd8e8), landed against #696:

- The exclusion policy resolves at the `index.ts` composition boundary rather than inside `createSubagentSession` itself, so the assembly factory stays free of policy and gains no new dependency-bag field.
- The child's filtered settings view is built via `SettingsManager.fromStorage` over a read-only filtering storage, rather than the `Proxy` over `SettingsManager` — avoids forwarding a growing SDK surface through a hand-rolled `get` trap.
- The filtered view reaches only the resource loader; the child session keeps its real, file-backed `SettingsManager`, narrowing the blast radius.
- Added a `fix:` for a data-loss path this PR's own `snapshot()` handling avoided incidentally but did not name: `/subagents:settings` rewrites the whole project settings file, so an unrelated setting change would have silently erased a hand-edited `excludedExtensionPackages`.
- `test/session/package-exclusions.test.ts` exercises a real `SettingsManager.fromStorage(...)`, so the tests pin Pi's actual `extensions: []` disable semantics rather than a stub.

The design rationale is written up in `docs/decisions/0002-extensions-on-a-minimal-core.md` (an ADR amendment) and the shipped plan/retro under `packages/pi-subagents/docs/`, if useful context for future contributions.

Every implementation commit carries a `Co-authored-by: leipeng <leipeng950504@gmail.com>` trailer. Closing this in favor of the merged reimplementation — thanks again for surfacing the problem and the working prototype.

PR #702, opened 2026-08-05
fix(pi-permission-system): export child detector and duplicate-author…

### LINKED ISSUE #699: pi-permission-system: Sibling authorizer extensions cannot detect registered child sessions → spurious "already registered" on every subagent start
### Package

pi-permission-system

### What do you want to change?

Export a public API so sibling extensions using `registerAuthorizer` can detect that they are running inside an in-process subagent child, and/or reliably identify the duplicate-registration error that results.

Concretely, one or both of:

**Option A: export a public child-context detector**

```ts
// public.d.ts
export function isRegisteredSubagentChild(ctx: Pick<ExtensionContext, "sessionManager">): boolean;
```

Backed by the existing `SubagentSessionRegistry` — no new infrastructure, just exposing what's already there.

**Option B: export a typed duplicate-registration error**

Currently `AuthorizerRegistry.register()` throws a bare `new Error(...)` with no `code` or subclass. Export a typed error so consumers can catch it reliably instead of matching the message string:

```ts
// public.d.ts
export class AuthorizerAlreadyRegisteredError extends Error {
  readonly code: "AUTHORIZER_ALREADY_REGISTERED" as const;
  readonly linkName: string;
}
```

Options A and B are complementary — A lets extensions skip the call entirely, B lets them handle the error reliably if it still occurs.

### Why?

`PermissionServiceLifecycle.activate()` suppresses `publishPermissionsService()` for registered children (#302), but `emitReadyEvent(this.events)` runs unconditionally:

```ts
// src/service-lifecycle.ts
activate(ctx: ExtensionContext): void {
  if (!this.detection.isRegisteredChild(ctx)) {
    publishPermissionsService(this.service);
  }
  emitReadyEvent(this.events); // ← unconditional
}
```

So every in-process child emits `permissions:ready`. Sibling authorizer extensions subscribe to `PERMISSIONS_READY_CHANNEL` and call `getPermissionsService()` in their handler. In a child, `getPermissionsService()` returns the **parent's** service (the child never published its own). The child calls `service.registerAuthorizer("my-link", ...)` on the parent's service, which already owns that name → `AuthorizerRegistry.register()` throws `An authorizer is already registered for 'my-link'.`

The detection capability exists internally — `isRegisteredSubagentChild(ctx, registry)` in `src/authority/subagent-context.ts`, backed by the process-global `SubagentSessionRegistry` (`Symbol.for("@gotgenes/pi-permission-system:subagent-registry")`). But neither is exported from `public.d.ts`.

The only "official" signal available to extensions is `permissions:ready`, which fires for both parent and child. There is no `hasAuthorizer(name)` query, no service/session identity, no child-context detector, no typed duplicate-registration error, and no error `code`.

**Impact:** benign but noisy. The child reuses the parent's authorizer (same service, same link), so authorization still works. But every in-process subagent start emits a warning per authorizer extension (2× if the extension registers from both `session_start` and `permissions:ready`). And because a genuine stale-registration failure (e.g. `/reload` dispose glitch) produces the same error, extensions cannot safely suppress the duplicate-registration warning without hiding real lifecycle bugs.

`@gotgenes/pi-permission-model-judge` has the same registration pattern (`session_start` + `permissions:ready`, no child detection) and is affected the same way.

### Option A implementation

1. In `src/authority/subagent-context.ts`, `isRegisteredSubagentChild` already takes `(ctx, registry)`. Add a thin wrapper that resolves the registry from `getSubagentSessionRegistry()` so callers don't need to:

   ```ts
   export function isRegisteredSubagentChild(ctx: SubagentDetectionContext): boolean {
     return isRegisteredSubagentChild(ctx, getSubagentSessionRegistry());
   }
   ```

   (Or export the two-arg form plus `getSubagentSessionRegistry`.)

2. Re-export from `public.d.ts`.

3. Extension usage:

   ```ts
   pi.events.on(PERMISSIONS_READY_CHANNEL, (_event, ctx) => {
     if (isRegisteredSubagentChild(ctx)) return; // child reuses parent link
     tryRegister();
   });
   ```

### Option B implementation

1. In `src/authority/authorizer-registry.ts`, replace the bare `throw new Error(...)` with a typed error:

   ```ts
   export class AuthorizerAlreadyRegisteredError extends Error {
     readonly code = "AUTHORIZER_ALREADY_REGISTERED" as const;
     constructor(readonly linkName: string) {
       super(`An authorizer is already registered for '${linkName}'.`);
       this.name = "AuthorizerAlreadyRegisteredError";
     }
   }
   ```

2. Re-export from `public.d.ts`.

3. Extension usage:

   ```ts
   try {
     dispose = service.registerAuthorizer(LINK_NAME, authorize);
   } catch (e) {
     if (e instanceof AuthorizerAlreadyRegisteredError) {
       // benign — child reuses parent's link
     } else {
       throw e; // or warn
     }
   }
   ```

### Option C (alternative): guard `emitReadyEvent`

Guard `emitReadyEvent` behind the same `isRegisteredChild` check so children don't emit `permissions:ready` at all. This changes the current event contract (children currently do emit), so it may affect other consumers. Less preferable than A/B.

### Reproduction

1. Install `@gotgenes/pi-permission-system` + `@gotgenes/pi-subagents` + `@gotgenes/pi-permission-model-judge`
2. Name the model-judge link in `authorizerChain`
3. Spawn an in-process subagent
4. Observe the duplicate-registration warning in console output

kuoruan:
…izer guard

Expose isRegisteredSubagentChild as a public API and add cross-extension-safe duplicate-registration handling via isAuthorizerAlreadyRegisteredError.

Keep duplicate registration typed internally while avoiding public instanceof coupling across isolated extension module instances.

Closes #699

### DIFF
--- a/packages/pi-permission-system/docs/cross-extension-api.md
+++ b/packages/pi-permission-system/docs/cross-extension-api.md
@@ -78,6 +78,16 @@ interface PermissionsService {
     toolName: string,
     extractor: (input: Record<string, unknown>) => string | undefined,
   ): () => void;
+
+  /**
+   * Register a named live-authority link consulted by `authorizerChain`.
+   * Returns a disposer.
+   * Throws a duplicate-registration error when the name is already present.
+   */
+  registerAuthorizer(
+    name: string,
+    authorize: Authorizer["authorize"],
+  ): () => void;
 }
 ```
 
@@ -262,6 +272,49 @@ const dispose = permissions.registerToolAccessExtractor("ffgrep", (input) =>
 
 Registration rules mirror `registerToolInputFormatter`: one extractor per tool name (a second `register` for the same name throws), and the returned disposer is identity-guarded.
 The extractor must not throw — guard your parsing and return `undefined` on anything unexpected.
+
+#### `isRegisteredSubagentChild` and `isAuthorizerAlreadyRegisteredError`
+
+`registerAuthorizer` callbacks receive the current ask details, the narrow `PermissionQuery`, the shared `AuthorizerLog`, and an optional `signal` (`ctx.signal`) as the fourth argument.
+Use `signal` to cancel long-running review work (for example, a model call) promptly when the user interrupts the session.
+
+When an extension registers authorizers from a `permissions:ready` listener, in-process subagent children reuse the parent service instance.
+A child can therefore hit duplicate registration if it calls `registerAuthorizer` for a link name the parent already owns.
+
+Use the public child detector to skip duplicate registration in children:
+
+```typescript
+import { isRegisteredSubagentChild } from "@gotgenes/pi-permission-system";
+
+pi.events.on(PERMISSIONS_READY_CHANNEL, (_event, ctx) => {
+  if (isRegisteredSubagentChild(ctx)) {
+    return;
+  }
+  registerAuthorizer();
+});
+```
+
+And/or use the exported type guard when a duplicate slips through:
+
+```typescript
+import {
+  getPermissionsService,
+  isAuthorizerAlreadyRegisteredError,
+} from "@gotgenes/pi-permission-system";
+
+try {
+  getPermissionsService()?.registerAuthorizer("model-judge", authorize);
+} catch (error) {
+  if (isAuthorizerAlreadyRegisteredError(error)) {
+    // Benign duplicate in child context.
+  } else {
+    throw error;
+  }
+}
+```
+
+Duplicate-registration errors carry a stable `code` (`"AUTHORIZER_ALREADY_REGISTERED"`) and the duplicate `linkName`.
+`isAuthorizerAlreadyRegisteredError` checks that shape directly, so handling stays reliable across per-extension module isolation.
 
 #### Subagent session registration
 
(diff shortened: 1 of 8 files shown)


gotgenes:
Thanks @kuoruan — the trace here (the duplicate-registration throw, and the SDK signature check on `permissions:ready`) is what started the investigation that produced ADR 0012.

This PR's proposals are superseded by the settled contract rather than merged: the exported child detector is process-specific by construction (an own-process subagent's registration succeeds silently into a registry nothing reads — the same defect, different symptom), and the typed duplicate-registration error is deferred — under the contract a remaining duplicate is a genuine author bug that should surface raw, and a `code` can land additively later if a real consumer need appears.

The shipped fix is session-keyed service publication: every node (root session or in-process subagent child) now publishes its own service under its own session id, so a sibling extension registers into the node's own registries instead of reaching the parent's. Landed on `main` in 4efba3f483fbd1b7aaea495bd8dd94b138187d51 (issue #699); see [ADR 0012](https://github.com/gotgenes/pi-packages/blob/main/packages/pi-permission-system/docs/decisions/0012-cross-node-extension-contract.md) for the full deliberation.

Closing as superseded — thanks again for the report that started this.

PR #705, opened 2026-08-05
fix(pi-subagents-worktrees): `cleanupWorktree` silently discards work when the fallback commit fails

### LINKED ISSUE #704: `cleanupWorktree` silently discards work when the fallback commit fails
### Package

pi-subagents

### Package version

v0.2.3

### What happened?

(not actually a "pi-subagents" bug, but I couldn't pick [pi-subagents-worktrees](https://github.com/gotgenes/pi-packages/tree/main/packages/pi-subagents-worktrees) in the above package picker which is the actual problematic package)

`cleanupWorktree` in `src/worktree.ts` permanently destroys an agent's uncommitted work if the safety-net commit it runs (`git add -A` + `git commit`) is rejected by a repo's pre-commit hook, or some other error occurs.

This happens with any hook that mutates files and then exits non-zero on its first pass - a very common formatter/linter pattern (`prettier --write`, `eslint --fix`, `rumdl-fmt`, etc.), not an edge case.

When git commit fails, `execFileSync` throws, the surrounding catch block force-removes the worktree (`git worktree remove --force`) and returns `{ hasChanges: false }` - as if nothing had changed. Since the commit never succeeded, the changes were never written to the object database either, so nothing is recoverable: no dangling commit, no dangling blob, `git fsck` finds nothing. The agent's work is silently and irrecoverably lost.

### Expected behavior

If an error occur during commit/cleanup, leave the worktree in place and return an error message to the agent such that it or the user can resolve the issue

### Steps to reproduce

1. In a repo, add a pre-commit hook that mutates a file and then exits non-zero (simulates a formatter/linter that auto-fixes but still fails the commit):
    ```sh
      #!/bin/sh
      echo "mutated" >> mutated-by-hook.txt
      git add -A
      exit 1
    ```
 2. Create a worktree and add an uncommitted file to it:
    ```sh
      git worktree add --detach /tmp/wt HEAD
      cd /tmp/wt
      echo "agent work" > new-file.txt
    ```
 3. Call `cleanupWorktree(cwd, worktree, "description")` on this worktree (or run its exact underlying commands: `git add -A then git commit -m "..."`).

### Additional context

_No response_

AndersBennedsgaard:
On commits in the worktree that fails for any reason (e.g. a failing pre-commit hook) now leaves the worktree in place and returns an error message.

Close #704

### DIFF
--- a/packages/pi-subagents-worktrees/README.md
+++ b/packages/pi-subagents-worktrees/README.md
@@ -44,6 +44,7 @@ An agent type not in `worktreeAgents` runs in the parent working directory, exac
 - A child whose agent type is listed gets a fresh detached worktree at `HEAD` before it runs.
 - When the child finishes with no changes, the worktree is removed.
 - When the child finishes with changes, they are committed to a branch (`pi-agent-<id>`), and the child's result gains a note: `Changes saved to branch \`<branch>\`. Merge with: \`git merge <branch>\``.
+- If cleanup fails partway through (e.g. a pre-commit hook rejects the safety-net commit), the worktree is **left in place** rather than removed, and the child's result gains a note: `Worktree cleanup failed and was left in place for manual recovery at \`<path>\`: <error>`.
 - If worktree creation fails for an opted-in agent (not a git repo, no commits yet, or `git worktree add` fails), the child run **fails** with an explanatory error rather than silently running unisolated.
 
 ## Migrating from `isolation: "worktree"`
--- a/packages/pi-subagents-worktrees/docs/plans/0704-preserve-worktree-on-cleanup-failure.md
+++ b/packages/pi-subagents-worktrees/docs/plans/0704-preserve-worktree-on-cleanup-failure.md
@@ -1,0 +1,206 @@
+---
+issue: 704
+issue_title: "cleanupWorktree silently discards work when the fallback commit fails"
+---
+
+# Preserve the worktree when cleanup fails partway
+
+## Problem Statement
+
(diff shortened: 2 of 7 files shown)


gotgenes:
Thanks for the clear diagnosis and reproduction, @AndersBennedsgaard — this confirmed a real data-loss bug in `cleanupWorktree`.

We ended up adopting the capability with a simplified design rather than merging this PR directly. The core fix (preserve the worktree instead of force-removing it on cleanup failure) matches what you proposed here. We also added a retry of the rescue commit with `--no-verify` (re-staging first) for the common trigger case — a formatter or linter hook that rewrites files and then exits non-zero — so that case now lands the work on a branch instead of just preserving the worktree. And we modeled the result as a discriminated union rather than adding optional fields to the existing shape.

Implemented in:

- 7411088c8a7a774c2e48627b8a83afdc9633c821 — extract git helpers from `cleanupWorktree`
- e1baca458e152d63270e2d9679589514ded295f9 — model cleanup result as a discriminated union
- 4a3f8e205e62db4bd7b664507f72a81f68a9029e — preserve the worktree when cleanup fails
- b9d17cbc1dc9917e3ad4129d7f21f1799f366b78 — retry the rescue commit with `--no-verify`
- 60ec8503ae8f71dee346b5d5238d7f05c7a8fe29 — document the new behavior in the README

Every one of those commits carries a `Co-authored-by` trailer crediting your diagnosis. Closing this in favor of that implementation, but the credit is yours — much appreciated.

PR #716, opened 2026-08-10
feat(pi-permission-system): improve permission dialog readability

Syndlex:
## Problem

Permission dialogs currently show everything as a single long sentence, which is hard to scan quickly:

```
Current agent requested bash command 'cd /home/user/repo' which references path(s) outside working directory '/home/user/project': /home/user/repo. Allow this external directory access?
```

This makes it difficult to:
- Identify at a glance what tool/command is being requested
- See the full command when it is long (it gets truncated)
- Distinguish the key facts (what, where, why) from boilerplate text

## Solution

Restructure permission prompts into aligned `key : value` lines, one fact per line.

### bash prompts
**Before:**
```
Current agent requested bash command 'cd /repo && git commit -m "msg"' (matched '*'). Allow this command?
```
**After:**
```
bash: cd /repo
  full    : cd /repo && git commit -m "msg"
```

### external_directory prompts
**Before:**
```
Current agent requested tool 'edit' for path '/home/user/.pi/agent/npm/...' outside working directory '/home/user/repo'. Allow this external directory access?
```
**After:**
```
tool     : edit
path     : /home/user/.pi/agent/npm/...
cwd      : /home/user/repo

⚠️  EXTERNAL DIRECTORY — allow access?
```

### tool prompts
Full JSON input is now pretty-printed instead of truncated to 200 chars.

## Changes

- **`permission-prompts.ts`**: `formatAskPrompt()` now returns structured multi-line strings instead of prose sentences. The generic `(matched '*')` fallback qualifier is suppressed since it adds no information.
- **`external-directory-messages.ts`**: Both `formatExternalDirectoryAskPrompt()` and `formatBashExternalDirectoryAskPrompt()` restructured into aligned key:value lines with a prominent `⚠️  EXTERNAL DIRECTORY` header.
- **`permission-prompt-component.ts`**: `renderDecision()` now renders message lines with contextual theme colors — `warning` for ⚠️ lines, `muted` for labels in `key : value` lines.

## Related issues

Relates to #654 (contextual explanations in permission prompts)

### DIFF
--- a/packages/pi-permission-system/src/authority/permission-prompt-component.ts
+++ b/packages/pi-permission-system/src/authority/permission-prompt-component.ts
@@ -248,7 +248,26 @@ class PermissionPromptComponent implements Component {
   }
 
   private renderDecision(): string[] {
-    const lines = [this.theme.fg("accent", this.title), this.message, ""];
+    const lines = [this.theme.fg("accent", this.title), ""];
+    // Render each line with contextual color:
+    // - lines with ⚠️  → warning color
+    // - "key : value" label lines → key in muted, value in text
+    // - empty lines → as-is
+    for (const line of this.message.split("\n")) {
+      if (line.includes("⚠️")) {
+        lines.push(this.theme.fg("warning", line));
+      } else if (line.includes(" : ")) {
+        const colonIdx = line.indexOf(" : ");
+        const label = line.slice(0, colonIdx);
+        const value = line.slice(colonIdx + 3);
+        lines.push(
+          this.theme.fg("muted", label) + this.theme.fg("muted", " : ") + value,
+        );
+      } else {
+        lines.push(line);
+      }
+    }
+    lines.push("");
     for (const key of OPTION_ORDER) {
       const label = key === "s" ? this.config.sessionLabel : OPTION_LABELS[key];
       const selected = this.state.highlightedKey === key;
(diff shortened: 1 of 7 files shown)


unrelentingfox:
Hi, looks like we have the same problem with two different approaches: https://github.com/gotgenes/pi-packages/pull/738

gotgenes:
Thanks for this — the readability problem you identified (a wall of prose with no way to scan what actually matters) was real, and the aligned `key : value` layout you proposed is close to what landed.

I implemented it as a *render* rather than an assembler change: gates now emit a structured payload, and the dialog renders it under a row budget and per-field width cap (packages/pi-permission-system/src/presentation/dialog-renderer.ts), rather than the assembler pre-formatting a string. That distinction mattered because the assembler version would have made the review log persist unbounded, unredacted tool input as a side effect of the readability fix — [ADR 0011](https://github.com/gotgenes/pi-packages/blob/main/packages/pi-permission-system/docs/decisions/0011-prompt-presentation-contract.md) has the reasoning if you're curious.

Landed in bcc61a3fa79c6a0f31cabe0328651b61b42b4f52 with a co-authorship credit in the commit, and closing this as superseded rather than rebasing it onto the new shape. Thank you for the idea and the writeup — much appreciated.

PR #729, opened 2026-08-13
feat(pi-subagents): add compact Ctrl+O-expandable rendering for get_subagent_result

### LINKED ISSUE #636: Add compact, Ctrl+O-expandable rendering for `get_subagent_result`
### Package

pi-subagents

### What do you want to change?

Make `get_subagent_result` participate in Pi's native collapsed/expanded tool-result behavior.

By default, it should display a compact summary instead of rendering the complete plain-text report. Pressing Ctrl+O (`app.tools.expand`) should reveal additional result content, consistent with the existing `subagent` tool.

This should also apply when `verbose: true` is requested: the verbose conversation may remain part of the tool result, but it should not be dumped into the collapsed TUI view.

### Why?

`get_subagent_result` currently has no compact result presentation, so Pi falls back to displaying its complete text output inline.

The report can contain:

- Agent identity, type, and status
- Execution statistics
- Description
- The complete agent result
- The full conversation when `verbose: true`
- The transcript path

Large results can contain tens of kilobytes and hundreds or thousands of lines. Rendering all of this by default can fill the terminal and push the surrounding conversation out of view.

This is also inconsistent with the other subagent surfaces:

- `subagent` results are compact and expandable with Ctrl+O.
- Background completion notifications use a compact themed presentation.
- Only `get_subagent_result` displays the full raw result by default.

### How? (optional)

   - Add custom `renderCall` and `renderResult` behavior to the `get_subagent_result` tool,
 following the existing rendering pattern used by `subagent`.
   - Use `ToolRenderResultOptions.expanded` so the presentation responds to Pi's native Ctrl+O
 tool expansion.
   - In the collapsed state, render agent identity, status, description, execution statistics,
 and an optional short result preview.
   - In the expanded state, render additional result content and the verbose conversation when
 requested, while keeping the displayed output bounded.
   - Pass compact structured presentation metadata through the tool result details rather than
 parsing the formatted plain-text report.
   - Avoid duplicating the complete result or conversation in the details object; the full text
 should remain in the existing tool-result content.
   - Apply display-width-aware limits so extremely long individual lines cannot create an
 unbounded number of terminal rows.
   - Preserve the complete tool-result content delivered to the parent model and stored in the
 session.

rosingrind:
- Add GetResultDetails interface for custom rendering metadata
- Add renderCall function showing agent ID and operation type
- Add renderResult function with collapsed/expanded states
- Collapsed shows agent identity, status, stats, description (no result)
- Expanded shows truncated result preview + verbose/transcript notes
- Never show verbose conversation in collapsed state
- Apply display-width-aware limits (80 chars desc, 200 chars/line, 30 lines)
- Fix aborted status color to match result-renderer (error, not warning)
- Fix compaction glyph to use GLYPHS.compactions
- Add 11 unit tests for rendering behavior

Close #636

### DIFF
--- a/packages/pi-subagents/src/tools/get-result-tool.ts
+++ b/packages/pi-subagents/src/tools/get-result-tool.ts
@@ -1,15 +1,171 @@
+import type { AgentToolResult, ToolRenderResultOptions } from "@earendil-works/pi-coding-agent";
 import { defineTool } from "@earendil-works/pi-coding-agent";
+import { Text } from "@earendil-works/pi-tui";
 import { Type } from "@sinclair/typebox";
 import type { AgentConfigLookup } from "#src/config/agent-types";
+import type { SubagentStatus } from "#src/lifecycle/subagent";
 import { type AgentReport, formatAgentReport } from "#src/tools/get-result-report";
-import { formatLifetimeTokens, textResult } from "#src/tools/helpers";
+import { formatLifetimeTokens, textResultWithDetails } from "#src/tools/helpers";
 import type { Subagent } from "#src/types";
-import { formatDuration, getDisplayName } from "#src/ui/display";
+import { formatDuration, getDisplayName, type Theme } from "#src/ui/display";
+import { GLYPHS } from "#src/ui/glyphs";
+
+// ---- Local helpers ----
 
 // ---- Deps interfaces ----
 
 export interface GetResultToolManager {
 	getRecord(id: string): Subagent | undefined;
+}
+
+/**
+ * Metadata attached to get_subagent_result tool results for custom rendering.
+ * Mirrors AgentDetails in agent-tool.ts but scoped to get_subagent_result.
+ */
+export interface GetResultDetails {
+	id: string;
+	displayName: string;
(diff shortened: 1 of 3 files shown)


gotgenes:
Thanks for the PR, @rosingrind — the width-aware limits and the collapsed/expanded split here are exactly the right shape, and both landed.

I reimplemented this in-repo (this project ties every change to its own TDD cycle and plan/retro docs) rather than merging directly, but your design is credited as a `Co-authored-by:` trailer on the landing commit, 46dc3f0e98fa19821be68c52aaba9b326ba0fba4, alongside the issue reporter.

A couple of things came out differently after digging into how Pi actually consumes `details`:

- `details` is written verbatim into the session JSONL and never sent to the model, so I kept it to compact metadata only rather than duplicating the full result text there.
- The row bound needed to account for embedded newline/CR/vertical-tab/form-feed characters in a line, not just `String.length` — a line with an embedded newline still costs the terminal extra rows even under a width-based clip.

Closing this as superseded. Thanks again for putting in the work here!

Issue #710, opened 2026-08-10
pi-permission-system: Forwarded subagent permission prompts render unbounded tool input inline and push the parent transcript out of view


aoguai:
### Package

pi-permission-system

### Package version

24.0.0

### What happened?

When a `pi-subagents` child requests permission for a large tool invocation, the forwarded permission prompt renders the complete raw tool input inline in the parent Pi TUI.

In my case, a `scout` subagent requested permission for a `bash`/PowerShell command containing a long here-string with a generated Markdown report. The parent session displayed `Permission Required (Subagent)` and then rendered the full command body, wrapping it across a very large number of terminal rows.

As a result, the permission prompt occupied almost the entire visible transcript area and pushed the earlier parent-session output out of the viewport. It looked as though the previous scrollback had disappeared, although the main problem appears to be an unbounded inline approval prompt rather than confirmed transcript data loss.

The forwarded request itself appears to work correctly. The issue is that an unbounded raw command payload is displayed as ordinary inline approval text.

### Expected behavior

Forwarded subagent permission prompts should remain bounded even when the pending tool invocation contains a very large argument.

I would expect the permission UI to show a compact summary by default, for example:

- requesting subagent name;
- tool name;
- a short command preview;
- total character count and line count; and
- the available approval choices.

The full tool input should be available only through an explicit expand/details action, or inside a height-limited and scrollable preview.

A large forwarded permission request should not consume the parent transcript viewport or make prior output difficult to inspect while approval is pending.

### Steps to reproduce

1. Install and enable both `pi-subagents` and `@gotgenes/pi-permission-system`.
2. Use Pi in interactive TUI mode with permission prompts enabled.
3. Start a subagent that is required to ask the parent session for permission before running `bash`.
4. Ask the subagent to execute a command with a large argument, for example a PowerShell here-string that writes a long Markdown report to a file.
5. Wait for the child permission request to be forwarded to the parent session.
6. Observe that the parent displays `Permission Required (Subagent)` followed by the complete, wrapped command body inline.

The issue becomes more obvious as the embedded content grows to hundreds or thousands of lines.

### Additional context

Environment:

- Package: `@gotgenes/pi-permission-system`
- Package version: `24.0.0`
- Pi Coding Agent: `0.84.1`
- pi-subagents: `0.45.1`
- OS: Windows
- Shell: PowerShell 7
- Pi TUI mode: `fullscreen`
- `terminal.clearOnShrink`: `false`

[...]

PR #738, opened 2026-08-14
Highlight the flagged command, path, or target in TUI permission prompts

unrelentingfox:
## What

A permission ask packs the agent name, the flagged element, the matched pattern, and the full command into one long sentence, so the part that needs review is easy to miss on a quick scan.
This change paints the flagged element in the theme's warning color while the TUI dialog renders.

Bash external-directory ask. /etc/hosts is painted in both the quoted command and the disclosure list, while /etc/hostsbackup stays plain.

The screenshot above also shows the whole-token rule: `/etc/hosts` paints where it stands alone, and the same text inside `/etc/hostsbackup` stays plain, so a path never splits mid-token.

### Bash command

The flagged sub-command paints in both the quoted sub-command and the full command, while the matched pattern and the trailing shell operator stay plain.

Bash command ask with the flagged sub-command painted twice and the matched pattern left plain.

### Tool path

The whole path token paints, including hyphens and the file extension.

Tool path ask with the whole path token painted.

### Tool external directory

The requested path paints while the working directory and the tool name stay plain.

Tool external-directory ask with the requested path painted and the working directory left plain.

### MCP target

The requested target paints and the matched pattern stays plain.

MCP target ask with the requested target painted and the matched pattern left plain.

## How

`LocalUserAuthorizer` supplies the highlight target to the dialog, and the inline prompt paints every whole-token occurrence in both the decision and denial-reason views.
A short command such as `ls` stays plain inside `lsof` and `/usr/bin/lsblk`.

Bash command asks highlight the flagged sub-command.
Bash path, bash external-directory, tool path, and tool external-directory asks highlight the offending path.
MCP asks highlight the requested target.
Forwarded subagent asks derive the target from the `accessIntent` the child already sends, because they carry no command.
That last one needs no wire-format change, and a child that sends no intent keeps the plain rendering.

This adds a `highlightText` field to `PromptPermissionDetails`, modeled on the existing `sessionLabel` field that no logger reads.

Highlighting happens only while the TUI renders.
The `select`/`input` fallback receives the unmodified message, and `writeReviewEntry` uses an explicit field allowlist that omits the highlight text, so events, logs, persisted messages, and forwarded payloads keep the plain text.

## Scope

Forwarded highlighting covers bash asks only, gated on `accessIntent.surface`, so a forwarded path, tool, MCP, or skill ask keeps the plain rendering.
Skill asks do not highlight, because those gates supply no highlight target.
External-directory asks highlight one offending path instead of every uncovered path.
The whole-token rule treats `/`, `.`, and `-` as token characters by design, so a command that appears inside a longer path stays plain there.

## Verification

- `pnpm --filter @gotgenes/pi-permission-system run check` passed.
- `pnpm --filter @gotgenes/pi-permission-system run test` passed: 132 test files, 2782 tests.
- `pnpm exec eslint packages/pi-permission-system` passed.
- `pnpm --filter @gotgenes/pi-permission-system run verify:public-types` passed, including the packaged consumer type check.
- Each gate test asserts the painted output against the real formatter message, so a future change that makes a highlight target diverge from the rendered text fails a test instead of silently disabling the highlight.
- I exercised all five ask surfaces in a live TUI, which produced the screenshots above.
- Biome, rumdl, and fallow do not run on this host (their native binaries need a newer GLIBC), so CI gates those.


gotgenes:
Thanks for this — the whole-token highlighting is a nice touch, and the screenshots made the intent very clear.

I adopted it as part of bounding the same dialog to a row/width budget ([#710](https://github.com/gotgenes/pi-packages/issues/710)), so it landed as a render over a structured payload rather than a `highlightText` field on `PromptPermissionDetails`. The flagged element is now derived from the payload's own decision-relevant value (or the escaping paths, for a bash external-directory ask), so the highlight can't drift from the rendered text the way a separately-carried field could.

Landed in bcc61a3fa79c6a0f31cabe0328651b61b42b4f52 with a co-authorship credit in the commit, and closing this as superseded rather than rebasing it onto the new shape. Thank you for the idea and the screenshots — much appreciated.

PR #740, opened 2026-08-15
feat(pi-subagents): background-by-default + rich non-blocking status kick-back

flip-engineer:
# feat(pi-subagents): background-by-default + rich non-blocking status kick-back

## Motivation

Two friction points showed up repeatedly when orchestrating many subagents from a parent agent:

1. **Accidental full-duration blocking.** `subagent` without an explicit `run_in_background: true` runs in the foreground, which blocks the parent for the *entire* agent lifetime. The overwhelmingly common intent when delegating to a subagent is "fire and continue" — foreground execution is the exceptional case (interactive or strictly sequential work). Making background the default removes a whole class of stalled orchestration sessions, while `run_in_background: false` remains available as the explicit opt-in for foreground.

2. **Thin status on a running agent.** `get_subagent_result` on a still-running agent returned only *"Agent is still running. Use wait: true or check back later."* — no signal about what the agent is doing, how long it has run, or how much it has consumed. That opacity pushed parent agents toward `wait: true` (blocking) or blind re-polling. The kick-back now carries a live status snapshot so the parent can make an informed decision (keep waiting, steer, or collect later) **without blocking**.

### 1. `subagent` defaults to background execution
- `src/config/invocation-config.ts`: `runInBackground` default `false` → `true`.
- `src/tools/agent-tool.ts`: `run_in_background` parameter description updated to document the new default (foreground is now opt-in).

### 2. `get_subagent_result` returns a rich live-status kick-back for running agents
When the target agent is still running, the report body now includes (all sourced from the existing `Subagent` record — no new instrumentation):

- **Current activity** — actively running tool name(s) (`activeTools`), or a tail snippet of the in-flight assistant response text (`responseText`, last 200 chars), or "idle between turns"
- **Turns** — `turnCount` (with `maxTurns` when set)
- **Tool calls** — `toolUses` so far
- **Tokens** — lifetime `input` / `output` / `cacheWrite` so far
- **Context used** — `getContextPercent()`
- **Elapsed runtime** — existing duration formatting

Files: `src/tools/get-result-report.ts` (render + optional `AgentReport` fields), `src/tools/get-result-tool.ts` (populate fields; tool description rewritten to document non-blocking default and `wait: true` as the explicit blocking opt-in).

## Compatibility

- All new `AgentReport` fields are **optional** — existing consumers (`result-renderer.ts`, tests) are unaffected.
- `wait: true` behavior is unchanged; only the default (non-blocking) path is enriched.
- The `run_in_background` flip is a **behavioral default change**: callers that relied on implicit foreground execution must now pass `run_in_background: false`. This is called out prominently in the tool description.

## Validation

- All four modified files transpile cleanly (`bun build --no-bundle`).
- Every record field consumed was verified against the `Subagent` getters in `src/lifecycle/subagent.ts` (`lifetimeUsage`, `turnCount`, `activeTools`, `responseText`, `maxTurns`) and `subagent-state.ts`.
- Live-verified in a running pi session: the kick-back renders activity/turns/tool-calls/tokens/context/elapsed for an in-flight agent without blocking.

## Example kick-back (after)

```
Agent: 5051217f-6406-45d
Type: DSv4FlashAgent | Status: running | Tool uses: 58 | 1.7M token | Context: 7% | Duration: 756.9s (running)
Description: Full mailbox export via service.svc

Agent is still actively working — not waiting for it. Collect its result after the completion notification, or pass wait: true to block explicitly.
Current activity: running tool(s) bash
Turns: 51 | Tool calls: 58
Tokens so far: in 1646469 / out 36883 / cache-write 0
Context used: 7%
Elapsed: 756.9s (running)
```


gotgenes:
Thanks for this, and for the detail in the description — the motivation section made it easy to see what each half is doing.

I'd like to split it, because I can take one half and not the other.

## The status kick-back: yes, please

`get_subagent_result` returning *"Agent is still running. Use wait: true or check back later."* and nothing else is a real gap. The fields you chose — active tool, turn count, tool uses, tokens, context percent, elapsed — all come off the existing `Subagent` record, so it costs no new instrumentation, and making the new `AgentReport` fields optional keeps it non-breaking. I would merge that on its own.

## The background-by-default flip: no

This one runs into a stated non-goal:

> _A global run-mode default._
> Foreground or background is a per-invocation argument and a per-agent frontmatter key; a global flip changes every existing agent file at once.

`runInBackground` defaulting to `true` changes the behavior of every agent file already on disk, including ones whose authors never considered run mode. A caller relying on implicit foreground gets a silent behavior change that the tool description can document but not prevent.

Would you be willing to open the kick-back as its own PR, confined to `src/tools/get-result-report.ts` and `src/tools/get-result-tool.ts`? I'll review that one promptly.

PR #747, opened 2026-08-15
fix(pi-subagents): pass parentSession and invocation through SDK spawn path

### LINKED ISSUE #724: SDK-spawned agents (SubagentsService.spawn) never populate invocation, so they're permanently invisible in the background widget
We're using `@gotgenes/pi-subagents` 19.2.2 from an extension (not published yet) that dispatches every one of its subagents programmatically through `getSubagentsService().spawn(type, prompt, options)`, never by having the model call the `subagent` tool itself. That's deliberate: we want deterministic dispatch, bounded timeouts, and no dependency on the model choosing to make the right tool call.

The problem: none of these agents ever show up in the native widget or the status bar, even while they're actively running in the background.

Traced it down to the `invocation` field. `agent-widget.ts`'s `listBackgroundAgents()` is the sole gate for what the widget (and status line) render:

```ts
private listBackgroundAgents(): Subagent[] {
  return this.manager.listAgents().filter(record => record.invocation?.runInBackground === true);
}
```

The tool path (`spawn-config.ts` + `background-spawner.ts`) builds an `AgentInvocation` object and passes it through:

```ts
id = manager.spawn(params.snapshot, identity.subagentType, execution.prompt, {
  ...
  isBackground: true,
  invocation: execution.agentInvocation,
});
```

The SDK path (`service-adapter.ts`'s `SubagentsServiceAdapter.spawn()`) sets `isBackground: true` but never builds or passes an `invocation` field at all:

```ts
return this.manager.spawn(snapshot, type, prompt, {
  description,
  model,
  maxTurns: options?.maxTurns,
  thinkingLevel: options?.thinkingLevel,
  inheritContext: options?.inheritContext,
  bypassQueue: options?.bypassQueue,
  isBackground,
});
```

So `record.invocation` is `undefined` forever for any agent spawned this way, and it's filtered out of the widget permanently, not just until some later event.

Related context: #423 (widget self-driving from lifecycle events) touches the same file but doesn't cover this gap. That one is about where `setUICtx` gets captured, not about which agents pass the `invocation` filter in the first place.

Question, not really a bug report: would you take a small patch that has `spawn()` in `service-adapter.ts` build a minimal `invocation` snapshot (something like `{ runInBackground: isBackground }`) and pass it through, mirroring what the tool path already does? Or is this scoping intentional, maybe tied to ADR-0004 Decision A framing the widget as "background agents only," with SDK-spawned agents assumed to be a different kind of consumer expected to render its own UI?

We already built a small fallback widget on our side as a workaround, so we're not blocked here, just want to know if this is a gap worth closing upstream or if there's a supported way to opt an SDK-spawned agent into the native widget that we're missing.

Nithin745:
Closes #724.

`SubagentsServiceAdapter.spawn()`, the SDK path used when a consumer calls `getSubagentsService().spawn()` directly instead of going through the model's `subagent` tool (for example from a slash-command handler), never set `invocation` on the config it passes to `manager.spawn()`. The widget's roster filter checks `record.invocation?.runInBackground === true` (`agent-widget.ts:168`), so every agent spawned through the SDK path was filtered out of the background widget permanently, not just until some later event.

The same method also never set `parentSession`, unlike the tool path (`background-spawner.ts:34`), so a consumer relying on the parent/child session link had no resolvable parent session id or file for SDK-spawned children.

Both fixes bring the SDK path to parity with what `background-spawner.ts` already does for tool-invoked spawns: pass `invocation: { runInBackground: isBackground }`, and pass `parentSession: { parentSessionId: ctx.sessionManager.getSessionId(), parentSessionFile: ctx.sessionManager.getSessionFile() }`.

#724 proposed this exact shape: "would you take a small patch that has spawn() in service-adapter.ts build a minimal invocation snapshot (something like { runInBackground: isBackground }) and pass it through, mirroring what the tool path already does?" This PR is that patch, plus the parentSession gap found while touching the same function.

Two commits, each with its own test:
- pass parentSession through the SDK spawn path
- populate invocation for SDK-spawned agents

### DIFF
--- a/packages/pi-subagents/src/service/service-adapter.ts
+++ b/packages/pi-subagents/src/service/service-adapter.ts
@@ -41,7 +41,8 @@ export class SubagentsServiceAdapter implements SubagentsService {
   ) {}
 
   spawn(type: string, prompt: string, options?: SpawnOptions): string {
-    if (!this.runtime.currentCtx) {
+    const ctx = this.runtime.currentCtx;
+    if (!ctx) {
       throw new Error("No active session — cannot spawn agents outside a session.");
     }
 
@@ -58,6 +59,12 @@ export class SubagentsServiceAdapter implements SubagentsService {
       inheritContext: options?.inheritContext,
       bypassQueue: options?.bypassQueue,
       isBackground,
+      // Parity with the tool path (background-spawner.ts:34) so
+      // create-subagent-session.ts:241 publishes a resolvable parentSessionId.
+      parentSession: {
+        parentSessionId: ctx.sessionManager.getSessionId(),
+        parentSessionFile: ctx.sessionManager.getSessionFile(),
+      },
     });
   }
 
--- a/packages/pi-subagents/src/service/service-adapter.ts
+++ b/packages/pi-subagents/src/service/service-adapter.ts
@@ -59,6 +59,9 @@ export class SubagentsServiceAdapter implements SubagentsService {
       inheritContext: options?.inheritContext,
       bypassQueue: options?.bypassQueue,
       isBackground,
+      // Parity with the tool path (background-spawner.ts:34) so the native
+      // widget's roster filter (agent-widget.ts:168) picks up SDK spawns too.
+      invocation: { runInBackground: isBackground },
       // Parity with the tool path (background-spawner.ts:34) so
       // create-subagent-session.ts:241 publishes a resolvable parentSessionId.
       parentSession: {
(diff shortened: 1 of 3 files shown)


Nithin745:
Superseded upstream. Thanks for the fixes.

0731bad (released in 20.0.1) sets `parentSession` in `SubagentsServiceAdapter.spawn()` from `runtime.getSessionInfo()`, which is what the parentSession half of this PR was for. Reading it off the runtime accessor is a better route than the `ctx.sessionManager` reach used here.

The `invocation` half is obsolete rather than merged. 8e19739 made background-ness first-class on the record as `Subagent.isBackground` and pointed the widget filter at it, so the widget no longer consults `invocation`, and 4942e09 then dropped `invocation` from the workspace seam in 21.0.0. Passing it through the SDK door would now be the wrong mechanism.

Nothing in this diff is still needed. Closing.

Issue #423, opened 2026-06-17
pi-subagents: make the agent widget self-drive from lifecycle events


gotgenes:
## What

Make the agent widget drive its own update timer from lifecycle notifications instead of being driven by the core spawn tools.
The widget should start and stop its timer in response to started / created / completed notifications; the spawn tools should no longer call `ensureTimer` / `update` / `markFinished`.

## Why

Today the dependency arrow points the wrong way: the core spawn tools call into the widget to drive it.
For the widget to be a clean reactive consumer of the core — and for the spawn tools to shed their widget dependency — the widget must subscribe to lifecycle signals and manage its own timer.

## Proposed change

- `src/ui/agent-widget.ts` — start/stop the update timer in response to lifecycle notifications rather than external `ensureTimer` calls; own its `markFinished` bookkeeping off observed completions.
- `src/lifecycle/subagent-manager.ts` (observer) and `src/index.ts` — wire the widget as a lifecycle consumer.
- Spawn tools no longer call widget methods.
- UICtx capture stays in `ToolStartHandler`.

## Context

Fourth step of the Phase 18 roadmap (widget/tool decoupling track).
Depends on removal of the activity map from the spawn tools (Phase 18 Step 3).
Unblocks dropping the widget dependency from the `subagent` tool in the next step.

gotgenes:
Phase 18 Step 4 is complete.

Implemented in f3bdb388684262c65828c37fef5cf2720196adba

Behavioral change: The agent widget now self-drives from lifecycle notifications (a reactive consumer) rather than being driven by inbound calls from the spawn tools. No change to observable behavior — non-breaking.

- refactor: add CompositeSubagentObserver to fan out lifecycle notifications
- refactor: subscribe the widget to lifecycle notifications via a composite observer
- refactor: stop driving the widget from the spawn tools (removed ForegroundWidgetDeps, BackgroundWidgetDeps, AgentWidget.markFinished; narrowed AgentToolWidget to setUICtx; made ensureTimer private)

This unblocks Step 5 (#424 — dropping the widget dep from the subagent tool).

Issue #422, opened 2026-06-17
pi-subagents: delete AgentActivityTracker and ui-observer, drop the activity map from the core


gotgenes:
## What

Delete `AgentActivityTracker` and `ui-observer`, and remove the activity map from `SubagentRuntime` and the spawn tools.
After the readers have migrated to the record getters, the tracker, its observer, and the map threaded through the core are dead and can go.

## Why

The activity tier is the UI's live-streaming bookkeeping, but today it lives inside the core: a `Map<string, AgentActivityTracker>` sits on the core composition root (`SubagentRuntime`), and the spawn tools construct trackers, open a second session subscription, and populate or delete the map.
Removing it leaves the core with one session subscription per child and zero UI state on the runtime — the structural win of the Phase 18 spine.

## Proposed change

- Delete `src/ui/agent-activity-tracker.ts` and `src/ui/ui-observer.ts` (~145 LOC).
- Remove `agentActivity` from `SubagentRuntime` (`src/runtime.ts`).
- In `src/tools/foreground-runner.ts` and `src/tools/background-spawner.ts`, stop constructing trackers, calling `subscribeUIObserver` / `setSession`, and populating the map.

## Context

Third step of the Phase 18 roadmap (activity-tier disentanglement spine).
Depends on the reader migration (Phase 18 Step 2), which removes the last references to the tracker.
After this step the core no longer does the UI's streaming bookkeeping; the widget/tool decoupling steps follow.

gotgenes:
Implemented in db3ee9d6

Phase 18 Step 3 (activity-tier disentanglement): deletes the dead AgentActivityTracker and ui-observer modules (~145 LOC), removes the SubagentRuntime.agentActivity map, and strips tracker wiring from the spawn tools. Each child session now has a single session subscription (the core's record-observer); the runtime holds zero UI state.

- refactor: stop wiring activity trackers in the spawn tools (#422)
- refactor: drop the activity map from the runtime and agent tool (#422)
- refactor: delete AgentActivityTracker and ui-observer (#422)
- docs: mark Phase 18 Step 3 complete and remove activity-tier references (#422)

No breaking change, no user-visible behavior change — the activity tier was already write-only dead state after Step 2's reader migration (#421). Unblocks Phase 18 Steps 4–8 (#423–#427).

PR #748, opened 2026-08-15
fix(pi-subagents): capture widget UI context at session_start and widen SubagentRecord

Nithin745:
A command-driven SDK spawn path in a downstream consumer (dispatch through `SubagentsService.spawn()` with no model tool call in the loop) can run an entire session with the widget dark. In one recorded run, a slash-command dispatch produced 35 transcript entries over 31 minutes with zero user messages and zero tool-call-triggered UI context, so `AgentWidget` never received a `UICtx` and never rendered anything for the whole run, even after the record became eligible for the widget's roster filter.

That eligibility half is #724 (this repo's other open item from us, not part of this PR). This PR is the other half: getting `UICtx` into the widget at all when nothing ever triggers `tool_execution_start`.

#423 deliberately kept `UICtx` capture in `ToolStartHandler`, and the linked ADR (0004-reconsider-ui-direction.md) pins that as an invariant Phase 19 has to preserve: "the runtime holds zero UI state (#422), the widget is a reactive consumer with no inbound calls from core spawn tools (#423), ..." I think this change keeps both invariants rather than breaking them: `session_start` isn't a call from a spawn tool into the widget, it's the composition root (`index.ts`) wiring the widget as a lifecycle observer, the same role it already plays via `observer.add(widget)`. The runtime doesn't hold any UI state either way; `ctx.ui` still comes from the event, same as it does in `ToolStartHandler`. So the question I'd actually ask: is the tool-call gate protecting something specific about when a UICtx is safe to capture, or was `ToolStartHandler` just where a `ctx` happened to already be in hand at the time?

The mechanism isn't new. It's the same `setUICtx` method, called with the same `ctx.ui` object, the session's UI context singleton. `ToolStartHandler` is unmodified and still the only place `onTurnStart()` gets called, so turn-aging is unaffected. `setUICtx`'s existing `ctx !== this.uiCtx` guard makes the later `ToolStartHandler` call a no-op once `session_start` has already run, so there's no double-capture or reset. The diff is small and additive: `AgentWidget` construction moves a few lines earlier in the same file so the `session_start` handler can reference it, and one line calls `widget.setUICtx(ctx.ui)`.

One thing worth flagging rather than hiding: `finishedTurnAge` only increments through `onTurnStart()`, and `onTurnStart()` only fires from `tool_execution_start`. In a session where a dispatch runs to completion and no parent turn ever happens, a completed agent's row is seeded at age 0 and never ages out of the widget until the user's next tool-calling turn. This PR doesn't touch that. If you'd rather not carry that as a loose end, I can pair this with either a non-turn-based aging signal or a timer-based sweep, whichever fits the widget's existing self-driving model better.

If a push-at-`session_start` model isn't what you want, the same fix works as a lazily-pulled supplier instead: something like `getUICtx: () => UICtx | undefined` that the widget reads from when it needs it, rather than being handed the value at `session_start`. Happy to send that version instead if you'd prefer it over the push model in this diff. (I did consider capturing `UICtx` inside `service-adapter.ts`'s `spawn()` itself, so the SDK path would self-sufficiently carry its own UI context, but that pulls UI concerns into the spawn path and runs against the same invariants above, so I left that out.)

Separately from the UICtx question: this PR also widens `SubagentRecord` with `turnCount: number` and `activeTools: string[]` as required fields, plus an optional `outputFile?: string`. All three already exist on the live `Subagent`/`SubagentState` but were dropped by `toSubagentRecord()`'s allowlist. Making `turnCount`/`activeTools` required (not optional) is a breaking change for anyone outside this repo implementing `SubagentRecord` themselves, since existing implementations won't have those fields. If you'd rather they stay optional, or be shaped some other way, I can rework that part before this merges as is; wanted to call it out explicitly rather than bury it in the diff.

Two commits:
- widen `SubagentRecord` with `turnCount`/`activeTools` (the breaking change above)
- capture widget UI context at `session_start`

### DIFF
--- a/packages/pi-subagents/src/index.ts
+++ b/packages/pi-subagents/src/index.ts
@@ -157,14 +157,23 @@ export default function (pi: ExtensionAPI) {
     unpublishSubagentsService,
   );
 
-  pi.on("session_start", (event, ctx) => lifecycle.handleSessionStart(event, ctx));
-  pi.on("session_before_switch", () => lifecycle.handleSessionBeforeSwitch());
-  pi.on("session_shutdown", () => lifecycle.handleSessionShutdown());
-
   // Live widget: constructed after the manager (it polls listAgents()) and
   // registered as a lifecycle observer so it self-drives its update timer.
   const widget = new AgentWidget(manager, registry);
   observer.add(widget);
+
+  pi.on("session_start", (event, ctx) => {
+    // Capture the UI context here too, not just on the first tool call
+    // (ToolStartHandler below): a spawn path that never triggers a tool
+    // call from the parent agent loop otherwise leaves the widget with no
+    // UI context for its entire run. Same setUICtx() method, same ctx.ui
+    // reference; setUICtx()'s existing `ctx !== this.uiCtx` guard makes the
+    // later ToolStartHandler call a no-op once this has already fired.
+    widget.setUICtx(ctx.ui);
+    return lifecycle.handleSessionStart(event, ctx);
+  });
+  pi.on("session_before_switch", () => lifecycle.handleSessionBeforeSwitch());
+  pi.on("session_shutdown", () => lifecycle.handleSessionShutdown());
 
   // Grab UI context from first tool execution + clear lingering widget on new turn
   const toolStart = new ToolStartHandler(widget);
--- a/packages/pi-subagents/src/service/service-adapter.ts
+++ b/packages/pi-subagents/src/service/service-adapter.ts
@@ -124,11 +124,16 @@ export function toSubagentRecord(record: Subagent): SubagentRecord {
     startedAt: record.startedAt,
     lifetimeUsage: record.lifetimeUsage,
     compactionCount: record.compactionCount,
+    turnCount: record.turnCount,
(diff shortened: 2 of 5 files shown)


Nithin745:
Superseded upstream. Thanks for picking up both halves.

The UI-context half shipped as 82bd707 ("render the agent widget in sessions with no model tool call", closes #827, released in 21.0.2). `WidgetEventsHandler.handleSessionStart` now calls `widget.setUICtx(ctx.ui)` on `session_start`, and the `tool_execution_start` capture is gone entirely, so a session that never triggers a tool call no longer leaves the widget without a context for its whole run. That is the case this PR was written for, and removing the tool-call capture goes further than the additive registration proposed here.

The `SubagentRecord` half is settled by ADR 0005 (docs/decisions/0005-subagent-record-admission-policy.md, accepted 2026-08-29), which names this PR directly: `turnCount` and `outputFile` are admitted, `activeTools` is declined under rule 2 as momentary activity, with a reactive channel named as the right route for anyone who needs it. `isBackground` and `maxTurns` landed too. I have no consumer for `activeTools` on the snapshot, so I am not pursuing the declined part.

Closing.

Issue #768, opened 2026-08-18
pi-subagents: settings key to append extra tool names to agent-type allowlists (additive counterpart of excludedExtensionPackages)


Nithin745:
**Package:** pi-subagents

**What do you want to change?**

Add an opt-in `additionalTools` key to `subagents.json` (global and project, loaded the same layered way as every other key) that appends tool names to an agent type's resolved allowlist at spawn time:

```json
{ "additionalTools": { "*": ["memory_search"], "general-purpose": ["skill_manage", "memory_add"] } }
```

Keys are agent-type names (case-insensitive, matching `resolveType`) or `"*"` for every type. The effective set is `getToolNamesForType(type) ∪ additionalTools["*"] ∪ additionalTools[type]`. Leaving the key out gives today's behavior, byte for byte. The recursion guard is unaffected.

**Why?**

Extension tools get registered inside every child, since children inherit the parent's extensions, but the per-type allowlist drops them anyway. The documented fix ("name the tool in `tools:`") doesn't have a good home for the default agents: overriding a default with a same-name `.md` file works, but the frontmatter can't express `isDefault`/`toolGuideline`, so the override drops that agent's guideline line from the `subagent` tool description and moves it out of "Default agents:". That's a real prompt-quality regression just to add one tool name, and a grant that should apply across agents turns into N hand-maintained prompt forks that drift on every release.

I read architecture.md's §Child tool selection and its line "the core does not widen this on the agent's behalf" before writing this up, and I think the proposal still respects that stance: it isn't inheriting whatever the parent's extensions happen to publish, it's an explicit, user-authored, default-off capability decision that's addressable per type (`"*"` is a choice you make, not a default you get). It also lines up with how `excludedExtensionPackages` landed (#696): a user-authored settings key rather than a provider seam, because the policy author is the user. This is basically the additive counterpart to that subtractive setting, and it's unrelated to the removed `memory:`/`skills:` preloading, since those injected content and this only touches the allowlist that §Tool selection already documents.

**How? (optional)**

[...]

PR #769, opened 2026-08-18
feat(pi-subagents): add additionalTools setting to append extra tool names to agent-type allowlists

Nithin745:
Refs #768

## What

Adds an opt-in `additionalTools` key to `subagents.json`.
It appends tool names to an agent type's resolved allowlist at spawn time, for both the `subagent` tool path and the SDK spawn path (`getSubagentsService().spawn()`).
Keys are agent-type names, case-insensitive, matching `resolveType()`, or `"*"` for every type.
The effective set is `getToolNamesForType(type) ∪ additionalTools["*"] ∪ additionalTools[type]`, deduplicated.
Leaving the key out of `subagents.json` is byte-for-byte identical to today's behavior.
The recursion guard (`excludeTools`) still strips this extension's own three tools even if someone lists them here.

## Why

Issue #768 has the full writeup.
Short version: extension tools are registered inside every child, since children inherit the parent's extensions, but the per-type allowlist drops them anyway.
The documented workaround, naming the tool in an agent's `tools:` frontmatter, has no good home for the three default agents.
Overriding a default with a same-name `.md` file works, but the frontmatter can't express `isDefault`/`toolGuideline`, so the override drops that agent's guideline line from the `subagent` tool description and moves it out of "Default agents:".

## How

`src/index.ts` gains three small exported helpers at the composition root: `sanitizeAdditionalTools`, `readAdditionalTools`, and `withAdditionalTools`.
`readAdditionalTools` runs a second, independently sanitized `loadLayeredSettings` pass over the same `subagents.json` file `SettingsManager` already loads, so it needs no changes to `settings.ts`'s own schema, sanitizer, or snapshot.
`withAdditionalTools` wraps the `AgentConfigLookup` handed to `subagentSessionDeps`, so `assembleSessionConfig`'s single `registry.getToolNamesForType(type)` call sees the extra grants on both spawn paths at once.
I picked this shape because it keeps the change to one file and mirrors how `loadLayeredSettings` is already meant to be used by extensions outside this package, but I'm not attached to it.
If you'd rather this ride through `settings.ts` properly (mirroring `excludedExtensionPackages`, including the snapshot round-trip so a project-level value survives a `/subagents:settings` write), I'm glad to rework it that way instead.

## Testing

`pnpm --filter @gotgenes/pi-subagents check && pnpm --filter @gotgenes/pi-subagents test` passes, including the new `test/additional-tools.test.ts` covering `sanitizeAdditionalTools`, `readAdditionalTools`, and `withAdditionalTools`.

### DIFF
--- a/packages/pi-subagents/src/index.ts
+++ b/packages/pi-subagents/src/index.ts
@@ -20,9 +20,10 @@ import {
   SettingsManager as SdkSettingsManager,
   SessionManager,
 } from "@earendil-works/pi-coding-agent";
-import { AgentTypeRegistry } from "#src/config/agent-types";
+import { type AgentConfigLookup, AgentTypeRegistry } from "#src/config/agent-types";
 import { loadCustomAgents } from "#src/config/custom-agents";
 import { InterruptHandler, SessionLifecycleHandler, ToolStartHandler } from "#src/handlers/index";
+import { loadLayeredSettings } from "#src/layered-settings";
 import { createChildLifecyclePublisher } from "#src/lifecycle/child-lifecycle";
 import { ConcurrencyLimiter } from "#src/lifecycle/concurrency-limiter";
 import { createSubagentSession, type SubagentSessionDeps } from "#src/lifecycle/create-subagent-session";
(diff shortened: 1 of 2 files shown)


Nithin745:
Closing this in favor of what I get from #775/#779: a settings-level key that widens a child's allowlist is the same class of problem as #612, even though the mechanism here was opt-in and explicit rather than an automatic union.

While building this I also checked whether overriding a default agent (e.g. `general-purpose`) via a same-name `.md` file has side effects. It does: `isDefault`/`toolGuideline` get silently dropped since the frontmatter parser doesn't read either. Folding the guideline text into the override's `description:` works around it fine though, so nothing else to raise there. Closing #768 too for the same reason.

Issue #827, opened 2026-08-29
pi-subagents: widget never renders in a session with no model tool call (UICtx captured only on tool_execution_start)


gotgenes:
`AgentWidget` can never render in a session where the model never calls a tool, because `UICtx` is captured at exactly one place:

```ts
// src/handlers/tool-start.ts:29
handleToolExecutionStart(_event: unknown, ctx: ToolStartCtx): void {
  this.widget.setUICtx(ctx.ui);
  this.widget.onTurnStart();
}
```

`AgentWidget.update()` opens with `if (!this.uiCtx) return;`, so with no `tool_execution_start` the widget is dark for the whole session — no widget, no status line — even for agents that pass its roster filter.

This is reachable today from a command-driven dispatch: a slash-command handler that calls `getSubagentsService().spawn(...)` runs agents with no model tool call anywhere in the loop. Reported downstream in #724 and PR #748, where one recorded run produced 35 transcript entries over 31 minutes with the widget never rendering.

## Why this is not settled by #423

#423 deliberately left `UICtx` capture in `ToolStartHandler`, and ADR-0004 pins the surrounding invariant: "the widget is a reactive consumer with no inbound calls from core spawn tools (#423)". The open question is whether the tool-call gate protects something specific about *when* a `UICtx` is safe to capture, or whether `ToolStartHandler` was simply where a `ctx` was already in hand.

## Verified SDK facts

Against the installed `@earendil-works/pi-coding-agent@0.79.1`:

- The TUI is up before extension init. `interactive-mode.js:498` calls `this.ui.start()` under the comment "Start the UI before initializing extensions so `session_start` handlers can use interactive dialogs"; extension binding and the `session_start` emission follow.
- `ctx.ui` is a stable per-session object reached through a getter on a per-event context (`runner.js:239`), so `session_start` and `tool_execution_start` see the same instance. `AgentWidget.setUICtx`'s existing `ctx !== this.uiCtx` guard therefore makes a later `ToolStartHandler` call a no-op.
- Headless binds `noOpUIContext` (`runner.js:87`), whose `setWidget`/`setStatus` are empty functions, so an early capture is safe there. `ctx.hasUI` is available to consult.

## Candidate approaches

1. Capture at `session_start` — the composition root wires the widget as a lifecycle observer, the same role it already plays via `observer.add(widget)`. This is PR #748's shape.
[...]

PR #850, opened 2026-08-30
Dispose AgentWidget during session shutdown (#849)

### LINKED ISSUE #849: pi-subagents: AgentWidget.dispose() is never called, so the widget is not torn down on session_shutdown
`AgentWidget.dispose()` is never called, so the agent widget survives `session_shutdown` with a live 80 ms interval and a registered widget/status entry.

## Current behavior

`AgentWidget` owns two resources that need tearing down: the `setInterval` started by `ensureTimer()`, and the `setWidget("agents", …)` / `setStatus("subagents", …)` registrations on the session's `UICtx`.
`dispose()` clears all of them, but nothing calls it — the method carries a `// fallow-ignore-next-line unused-class-member` comment so dead-code analysis stays quiet about it:

```console
$ rg -n 'widget\.dispose|AgentWidget.*dispose' packages/pi-subagents/src
(no call site)
```

`SessionLifecycleHandler.handleSessionShutdown()` unpublishes the service, clears the runtime session context, disposes notifications, aborts all agents, and disposes the manager.
It does not touch the widget.

The only other path that clears the interval is `clearWidget()`, which `update()` reaches once no agent is active or lingering — so in a normal session the timer does stop.
It is the shutdown path that has no teardown.

## Expected behavior

The widget is torn down on `session_shutdown` alongside the rest of the extension, the way `NotificationManager.dispose()` already is.

## Notes

Before #827, the interval could outlive a session a second way: `update()` returns early when no `UICtx` was ever captured, which makes `clearWidget()` unreachable — so a background spawn in headless mode, or before the first tool call, left the interval running until process exit.
#827 captures the `UICtx` at `session_start`, which closes that path.
The missing `dispose()` wiring is the remaining one.

Found while planning #827.
The fix is small, but it needs a decision on whether the widget joins `SessionLifecycleHandler`'s dependency set or gets its own `session_shutdown` registration beside the widget's other event wiring, so I'm filing it rather than folding it into that plan.

mikemikimike:
## Summary / Problem

Fixes #849.

`AgentWidget` owns a polling interval and UI registrations, but `session_shutdown` never called its existing `dispose()` method. A session could therefore retain the interval and widget/status entries until process exit.

## Changes

- Inject widget disposal into `SessionLifecycleHandler` and call it during shutdown while the UI context is still valid.
- Wire the composition root to dispose the live `AgentWidget`.
- Add lifecycle-order and timer/UI-registration regression tests.

## Tests

- `pnpm --filter @gotgenes/pi-subagents run check` — passed.
- `pnpm --filter @gotgenes/pi-subagents run test -- test/handlers/lifecycle.test.ts test/ui/agent-widget.test.ts` — 35 passed.
- `pnpm --filter @gotgenes/pi-subagents run test` — 1346 passed, 4 failed. The failures are existing Windows path-separator expectations in `test/session/session-dir.test.ts` and `test/composition-root.test.ts`.
- `pnpm --filter @gotgenes/pi-subagents run lint` — Biome and ESLint passed; the Markdown subcommand could not expand `*.md` on Windows.
- `git diff --check` — passed.

## Compatibility / Known limitations

The lifecycle constructor keeps a no-op default for callers that do not provide a widget. Full-suite and Markdown-lint limitations are environment-specific and are recorded above.

## Issue link

Fixes #849

### DIFF
--- a/packages/pi-subagents/src/handlers/lifecycle.ts
+++ b/packages/pi-subagents/src/handlers/lifecycle.ts
@@ -26,6 +26,7 @@ export interface LifecycleRuntime {
  * Constructor deps:
  * - `runtime` — owns session context state
  * - `manager` — manages agent lifecycle (clear, abort, dispose)
+ * - `disposeWidget` — unregisters the live widget and stops its timer
  * - `disposeNotifications` — tears down the notification system on shutdown
  * - `unpublishService` — unpublishes the SubagentsService symbol on shutdown
  */
@@ -35,6 +36,7 @@ export class SessionLifecycleHandler {
     private readonly manager: LifecycleManager,
     private readonly disposeNotifications: () => void,
     private readonly unpublishService: () => void,
+    private readonly disposeWidget: () => void = () => {},
   ) {}
 
   handleSessionStart(_event: unknown, ctx: unknown): Promise<void> {
@@ -48,15 +50,17 @@ export class SessionLifecycleHandler {
 
   // Cleanup order matters:
   // 1. Unpublish service — prevent new cross-extension calls
-  // 2. Clear session context — no more session state
-  // 3. Dispose notifications — silence nudges *before* the aborts that would
+  // 2. Dispose the widget — unregister UI state and stop its timer while its context is valid
+  // 3. Clear session context — no more session state
+  // 4. Dispose notifications — silence nudges *before* the aborts that would
   //    raise them: no parent run is active at shutdown, so a terminal
   //    transition delivers its nudge synchronously and Pi cannot recall it
-  // 4. Abort all agents — stop running and queued work
-  // 5. Dispose manager — final cleanup, awaited so each child's extensions get
+  // 5. Abort all agents — stop running and queued work
+  // 6. Dispose manager — final cleanup, awaited so each child's extensions get
   //    their `session_shutdown` before Pi tears the parent down (#709)
   handleSessionShutdown(): Promise<void> {
     this.unpublishService();
+    this.disposeWidget();
     this.runtime.clearSessionContext();
     this.disposeNotifications();
     this.manager.abortAll();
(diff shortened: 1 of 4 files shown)


gotgenes:
Thanks for this, @mikemikimike — your diagnosis was correct, and it's the one that shipped.

This repo reimplements adopted third-party changes through its own TDD cycle rather than merging them, so the fix landed as a9a7a6302880aa3b405585ca3784ff56eecc0477 with `Co-authored-by: mikemikimike <13286568797@163.com>`, released in `@gotgenes/pi-subagents` v21.2.1.

Adopted from your PR:

- The diagnosis and the shape of the fix — call `AgentWidget.dispose()` on `session_shutdown`.
- Your widget test's assertion set (`vi.getTimerCount()` → 0, `setWidget` and `setStatus` last called with `undefined`). That's the right pin, and it's what the shipped test asserts.

Two places where the implementation diverged, both worth recording:

1. **Ordering.** The PR calls `disposeWidget()` at position 2, before `abortAll()`. A spike measured that as undone whenever an agent is running or queued at shutdown: `Subagent.stopQueued()` calls `observer.onRunFinished()` inline, which drives `AgentWidget.update()` and re-registers the widget. The shipped teardown runs after the aborts instead.
2. **Placement and the defaulted dependency.** Instead of a `disposeWidget` parameter on `SessionLifecycleHandler`, the teardown became `WidgetEventsHandler.handleSessionShutdown()` — `WidgetEventsHandler` landed in #827 after your PR was written, and this keeps all three of the widget's host events in one module. The `= () => {}` default was also dropped: an unwired composition root should fail a test rather than silently no-op, which is the defect class this issue was about in the first place.

`dispose()` now also clears its `UICtx`, so `update()` returns at its first line afterward and disposal is final by construction rather than by call ordering.

Closing as shipped rather than merged, per that convention. Much appreciated.

Issue #812, opened 2026-08-25
pi-subagents: runtime-registered providers (e.g. pi-claude-bridge) unresolvable in child sessions


georgeharker:
### Package

pi-subagents

### Package version

19.3.5

### What happened?

Subagents cannot resolve any provider that was registered at **runtime** via `pi.registerProvider` — the provider is present in the parent session but missing in every child. With `pi-claude-bridge` as the active provider, spawning any subagent fails with:

```
Error: No API key found for claude-bridge.
```

Root cause: Pi ≥ 0.80.8 replaced `createAgentSession`'s `modelRegistry` option with `modelRuntime` (`CreateAgentSessionOptions` now exposes only `modelRuntime?: ModelRuntime`). The composition root in `src/index.ts` still passes only `modelRegistry`:

```ts
createSession: ({ sessionManager, resourceLoader, modelRegistry, ...rest }) =>
  createAgentSession({
    ...rest,
    sessionManager: sessionManager as SessionManager,
    resourceLoader: resourceLoader as ResourceLoader,
    modelRegistry: modelRegistry as SdkModelRegistry, // ignored by the SDK on >=0.80.8
  }),
```

Because the SDK ignores `modelRegistry`, each child builds a **fresh** runtime from config + auth, which contains none of the parent's runtime-registered providers, so the provider is unresolvable in the child.

### Expected behavior

A subagent inherits the parent's providers — including runtime-registered ones — so a subagent launched under `claude-bridge` resolves the model exactly as the parent does. `@tintinweb/pi-subagents` behaves this way because it forwards the parent's `modelRuntime` (shim in its `agent-runner.ts`).

### Steps to reproduce

1. Pi 0.84.3, with `pi-claude-bridge` installed and set as the default provider (registers a provider with `apiKey: "not-used"` at runtime), default model `claude-bridge/claude-opus-4-8`.
2. Spawn any subagent (e.g. the `subagent`/`Explore` tool).
3. The child turn fails with `No API key found for claude-bridge`.

Not bridge-specific: any extension that registers a provider via `pi.registerProvider` (rather than config/auth) reproduces it.

### Additional context

- Pi 0.84.3, macOS (Node 22+).
- Selecting a different model per child is orthogonal — `modelRuntime` is the provider pool, not the selection — so a haiku `Explore` under an opus parent is fine once the runtime is forwarded.
[...]

Issue #400, opened 2026-06-14
perf(pi-subagents): include parent system prompt in replace mode for KV cache reuse


gotgenes:
## Overview

Replace-mode subagents currently discard the parent system prompt entirely, substituting a thin two-line header:

```text
You are a pi coding agent sub-agent.
You have been invoked to handle a specific task autonomously.
```

This means replace-mode agents lose the core identity (`"You are an expert coding assistant operating inside pi..."`), tool usage guidelines, and AGENTS.md context that the parent carries — and they share no prompt prefix with the parent or with other subagents, defeating KV cache reuse.

## Context

@jeffutter noted in [#180 (comment)](https://github.com/gotgenes/pi-packages/issues/180#issuecomment-4653868095) that their replace-mode `researcher` agent's system prompt diverges from the parent almost immediately:

> I assumed it would have had the `You are an expert coding assistant operating inside pi, a coding agent harness. You help users by reading files, executing commands, editing code, and writing new files....` before the "you are a researcher" bit.

The `parentSystemPrompt` parameter is already passed into `buildAgentPrompt()` but replace mode ignores it completely.

## Proposed fix

Include the parent system prompt in replace mode, placed before the custom instructions — similar to how append mode now works after #180.
The replace-mode output would become:

```text
1. parentSystemPrompt (or genericBase)    ← SHARED, cacheable prefix
2. <active_agent name="${name}"/>         ← varies per agent
3. # Environment ...                      ← varies per runtime
4. config.systemPrompt                    ← custom instructions (full control)
```

This gives replace-mode agents the same KV cache prefix benefit that #180 introduced for append mode, and preserves the parent identity context that users expect.

The key difference from append mode remains: replace mode does not inject the `<sub_agent_context>` bridge or wrap custom instructions in `<agent_instructions>` tags — the custom prompt gets full control of the agent's instructions.

## Related

- #180 — reordered append-mode prompt for KV cache reuse (shipped in `pi-subagents-v6.18.3`)
- #360 — change custom agent default from replace to append

gotgenes:
Implemented in 1cc25cf0, shipped in 9c29ab2e.

- `perf!:` 1cc25cf0 — rewrite replace branch of `buildAgentPrompt()` to place `parentSystemPrompt` (or `genericBase` fallback) first, followed by the `<active_agent>` tag, env block, and `config.systemPrompt`
[...]

Issue #180, opened 2026-05-24
perf(pi-subagents): reorder append-mode system prompt to enable KV cache reuse


gotgenes:
## Overview

Subagent system prompts in append mode defeat LLM KV cache reuse because varying, agent-specific content precedes the large shared prefix.
The inherited parent system prompt (~8k tokens) is identical across all subagent invocations, but the LLM can never cache it because different tokens come first every time.

## Context

@jeffutter reported this in [tintinweb/pi-subagents#73 (comment)](https://github.com/tintinweb/pi-subagents/pull/73#issuecomment-4520786634).
They measured 8,333 shared tokens between the parent and a `scout` subagent, and noted that reprocessing those tokens on every spawn costs ~40 seconds on their hardware.
Our fork assembles prompts the same way in [`prompts.ts`](https://github.com/gotgenes/pi-packages/blob/main/packages/pi-subagents/src/session/prompts.ts), so we have the same problem.

## Current behavior

In append mode, `buildAgentPrompt()` assembles the system prompt in this order:

```text
1. <active_agent name="${name}"/>     ← VARIES per agent (cache-busting)
2. # Environment ...                  ← VARIES per runtime
3. <inherited_system_prompt>          ← SHARED (~8k tokens, cacheable)
4. <sub_agent_context>                ← SHARED (static)
5. <agent_instructions>               ← VARIES per agent
6. memory / skills                    ← VARIES
```

LLM KV caching works on **prefixes** — the cache is only reusable when the beginning of the prompt matches.
Since the varying `<active_agent>` tag and env block come *before* the 8k shared tokens, every subagent spawn reprocesses all of them from scratch.

## Proposed fix

Reorder the prompt so shared/stable content comes first:

```text
1. <inherited_system_prompt>          ← SHARED (~8k tokens, NOW CACHEABLE)
2. <sub_agent_context>                ← SHARED (static)
3. <active_agent name="${name}"/>     ← VARIES (after cached prefix)
4. # Environment ...                  ← VARIES
5. <agent_instructions>               ← VARIES per agent
6. memory / skills                    ← VARIES
```

This is a low-risk reordering:

- The `<active_agent>` tag just needs to be *present* in the system prompt for `pi-permission-system` to parse it; position doesn't matter.
- The env block provides context to the model; its position relative to the inherited prompt is not semantically significant.
- The inherited system prompt content serves as the "base" that gets specialized — placing it first is arguably more natural anyway.

[...]

Issue #89, opened 2026-05-20
Scaffold `@gotgenes/pi-colgrep` package


gotgenes:
Create the package structure for `@gotgenes/pi-colgrep` — a Pi extension that integrates [ColGrep](https://github.com/lightonai/next-plaid#colgrep) semantic code search as a tool available to the agent.

## Context

ColGrep is a fully local semantic code search CLI built on multi-vector ColBERT embeddings and tree-sitter parsing.
It combines regex filtering with semantic ranking, supports 25 languages, and runs entirely on the user's machine.
This package will expose ColGrep as a Pi tool that complements (not replaces) the built-in `grep`.

## Scope

Set up the package infrastructure following the monorepo conventions established by the existing packages.

### Package structure

- `packages/pi-colgrep/package.json` — standard fields (`type: "module"`, `engines.node: ">=22"`, `pi.extensions`, `publishConfig.access: "public"`, `catalog:` deps)
- `packages/pi-colgrep/tsconfig.json` — extends `../../tsconfig.base.json`
- `packages/pi-colgrep/LICENSE` — MIT
- `packages/pi-colgrep/CHANGELOG.md` — empty (release-please will own it)
- `packages/pi-colgrep/src/extension.ts` — minimal entry point that exports default function accepting `ExtensionAPI`
- `packages/pi-colgrep/README.md` — brief description, prerequisites (install colgrep), install instructions

### Monorepo integration

- Add to `pnpm-workspace.yaml` (already covered by the `packages/*` glob)
- Add entry to `release-please-config.json`
- Add entry to `.release-please-manifest.json`
- Verify `scripts/publish-released.sh` handles the new package
- Run `pnpm install` to wire up the workspace

### Verify

- `pnpm -C packages/pi-colgrep run check` passes (TypeScript)
- `pnpm -C packages/pi-colgrep run lint` passes (Biome + rumdl)
- CI runs cleanly with the new package in the workspace

gotgenes:
Implemented in c622d44

- feat: scaffold @gotgenes/pi-colgrep package (#89)
  Created the package infrastructure: package.json, tsconfig.json, LICENSE, CHANGELOG.md, AGENTS.md, README.md, and src/extension.ts (no-op entry point)
- chore: wire pi-colgrep into release and publish (#89)
  Added entries to release-please-config.json, .release-please-manifest.json, and scripts/publish-released.sh
- chore: update lockfile for pi-colgrep (#89)

[...]

PR #884, opened 2026-09-05
feat(pi-subagents): portable prompt inheritance for children whose provider re-homes the prompt

### LINKED ISSUE #883: pi-subagents: subagents on a Claude-Agent-SDK provider (pi-claude-bridge) get 400 "Third-party apps now draw from your extra usage" — the inherited parent prompt is the trigger
*Follow-up to #812 — the provider-inheritance fix that landed there is what makes this reachable: child sessions now embed the parent's assembled system prompt, and through a provider like pi-claude-bridge that prompt now travels into another harness's API requests.*

**TL;DR:** Since child prompts embed the **parent's assembled system prompt verbatim** (both `prompt_mode`s), any subagent running on a provider that forwards pi's system prompt to another harness — e.g. `pi-claude-bridge`, which appends the projected prompt to Claude Code's preset — now sends pi's base prompt text to Anthropic. Anthropic's subscription OAuth gate classifies that content as a **third-party app** and refuses the request unless the account has extra-usage credit. Parent and AskClaude on the same bridge, same token, same binary, pass. It is purely the inherited-prompt feature that trips it.

## Symptom

- pi-subagents **21.4.1**, pi-claude-bridge 0.7.0 (@anthropic-ai/claude-agent-sdk 0.2.141, Claude Code 2.1.141 and 2.1.260 both tested)
- A subagent whose model is the claude-bridge provider fails on its **first API call**:

```
API Error: 400 Third-party apps now draw from your extra usage, not your plan limits.
Add more at claude.ai/settings/usage and keep going.
```

- The **parent** on the same bridge provider: fine. **AskClaude called from inside the failing subagent**: fine. Same OAuth token, same machine, same process, same SDK spawn. Only subagent sessions fail.

## Root cause chain

1. `buildAgentPrompt()` places the parent's **entire assembled prompt** first in the child's system prompt, verbatim — in both `append` and `replace` modes (KV-cache rationale). `prompt_mode: replace` no longer avoids it; my repro agent has `prompt_mode: replace` and still fails.
2. claude-bridge projects a session's system prompt into Claude Code's `--append-system-prompt`. For the **parent**, its projection only carries the portable parts (context files, skills, custom/append) — pi's *base* prompt never reaches the API. That's why the parent is fine.
3. For a **child**, the inherited parent prompt arrives as ordinary `custom` text with no provenance. The bridge has an inheritance-matching safeguard (swap an exactly-embedded known prompt for its portable parts), but it requires the parent's capture to exist in the same in-memory map — and each session loads its own extension instance, so it can never match across the parent/child boundary. The base prompt flows through verbatim.
4. Anthropic's server scores the appended system text. Bisected to a **two-phrase trigger inside one line of pi's base prompt**:

   ```
   - When asked about: ... custom providers (docs/custom-provider.md) ... pi packages (docs/packages.md)
   ```

   - `"custom providers"` alone → passes. `"pi packages"` alone → passes. **Those two clauses alone, in an otherwise-empty append → 400, deterministically.**
   - `"banana packages"` instead of `"pi packages"` → passes. Dropping either phrase → passes. It's a cumulative scorer; the other eight doc-topic clauses ride near the threshold.

5. Classified as third-party app → draws from extra usage → no credit on the account → hard 400.

## Minimal repro (no pi, no bridge, ~30 lines)

Any Agent SDK query with `systemPrompt: { type: "preset", preset: "claude_code", append }` where the append contains both phrases:

```js
const append = "- When asked about: custom providers (docs/custom-provider.md), pi packages (docs/packages.md)";
// → 400 Third-party apps now draw from your extra usage...
const append = "- When asked about: custom providers (docs/custom-provider.md), banana packages (docs/packages.md)";
// → 200, normal plan usage
```

Everything else held constant (cwd, tools `[]`, in-process MCP custom-tools server with the subagent's exact 8 tools and descriptions, settings, extraArgs `strict-mcp-config`/`model`/`thinking-display`, effort, single-message history, OAuth auth).

## Ruled out (all tested, all innocent)

- Claude Code binary/version — 2.1.141 (SDK-bundled) and 2.1.260 (PATH) both pass and fail depending only on append content
- `cc_entrypoint=sdk-ts`, the `cc_version` fingerprint suffix (it's `sha256(salt + firstUserMsg[4,7,20] + version).slice(0,3)`, not a classification signal)
- MCP-served tool names/descriptions/schemas (`mcp__custom-tools__*`), `tools: []`, `--strict-mcp-config`, settings/claudeMdExcludes
- `<active_agent>` tag, env header, skills block, the "operating inside pi, a coding agent harness" identity line, cwd, model id, `[1m]`, effort, resume vs fresh session, prompt streaming
- Account/token state — AskClaude from the *same failing child session* passes

## Cross-check: @tintinweb/pi-subagents does not trip this

An independent implementation of the same extension (@tintinweb/pi-subagents 0.19.0) provides a control experiment on the same machine, token, and bridge:

- Its **`replace` mode is a thin standalone prompt** (active_agent tag + short sub-agent header + env + agent body — no parent prompt). Custom agents **default to replace** there. Subagents built this way pass Anthropic's gate — the pi base prompt never leaves the machine.
- Its **`append` mode inherits the parent prompt verbatim**, byte-identical-prefix rationale and all — and fails identically on the bridge.
- Instead of relying on the inherited base for context, it injects *specific* portable content into the child prompt (per-agent memory blocks, preloaded skill contents). That is "inherit the portable parts, not the harness base" — and it demonstrably stays under the classifier's threshold.

Same machine, same OAuth token, same bridge, same Claude Code binary: thin-standalone children pass, parent-base-inheriting children 400. The parent-prompt inheritance is the only variable that matters.

## Why this lands on pi-subagents

The byte-identical inherited prefix only pays off when the child talks to the same raw API as the parent. For a bridge-served child, the prompt is re-homed as an *append* onto another harness's preset — where inheriting the parent's full base is (a) cache-useless and (b) now actively toxic. The information a subagent actually needs from the parent prompt is the portable parts (context files, skills, persona/custom) — not pi's base harness instructions, which the host harness already supplies its own equivalent of.

## What I'd suggest (any one of these fixes it)

1. **Inherit only the portable parts.** Embed the parent's context-files/skills/custom projection rather than the assembled base. Keeps the intent of the inheritance change without hauling pi's harness text into foreign APIs. (This is effectively what @tintinweb/pi-subagents' replace+extras already does.)
2. **Expose provenance.** If pi/pi-subagents marks the inherited region (a field, or a stable boundary marker), downstream extensions like claude-bridge can strip it when projecting. claude-bridge's code already anticipates wanting exactly this.
3. **Make base-prompt inheritance opt-out.** A frontmatter key (`inherit_prompt: false` or similar) restoring a standalone child base, for users whose providers can't tolerate the inherited text.

(1) or (2) also restores the bridge's ability to keep prompt-cache alignment where it *does* work.

## Reference implementation (working, tested)

Option 1 + provider scoping is implemented and live-tested on my fork — `georgeharker/pi-packages`, branch **`feat/portable-prompt-inheritance`** (single commit `d75c678`, based on upstream `pi-subagents` 21.4.2):

- Frontmatter: `inherit_prompt: full | portable` per agent (wins over everything)
- Settings: `"promptInheritance": "portable"` (global) or `{ "providers": { "claude-bridge": "portable" } }` — the child's resolved model's provider picks the strategy, so bridge-served children go portable while same-API children keep the byte-identical cache prefix
- The portable identity is built from the parent's `before_agent_start.systemPromptOptions` (context files, custom/append prompts, added guidelines) — authoritative parts from pi's own event, no prompt-text heuristics. Skills stay un-inherited (the child loads its own catalogue); context files ride along because the child loader suppresses them.
- Default unchanged: `full` — no behavior change for anyone who doesn't opt in.

Verified end-to-end on the bridge (same machine/token): full-inheritance child → 400; portable child (either frontmatter or provider rule, built-in agent, no per-agent config) → succeeds on plan usage. Unit tests pin the frontmatter > provider > default resolution order.

## Note on the failure signal

The 400 only appears while the account has **zero extra-usage credit**. With credit added, the same classified request succeeds and bills to extra usage instead — the classification (and the cost) remains, only the error disappears. To verify the fix with credit on the account, check the rate-limit telemetry instead: classified requests report overage/extra-usage draw, fixed requests report plan windows.

Happy to supply the full bisect matrix, captured query dumps, and the SDK repro script.

georgeharker:
Fixes #883. Follow-up to #812.

## What

An opt-in **`portable`** prompt-inheritance strategy for agents whose provider re-homes the child's prompt into another harness (pi-claude-bridge projecting it onto Claude Code's preset as an append). The child's identity is built from the parent's `before_agent_start.systemPromptOptions` — context files, added guideline bullets, custom and append prompts — instead of the assembled prompt, so pi's harness base never leaves pi through that path.

Selection, per child: **agent frontmatter (`inherit_prompt`) > provider rule (`promptInheritance.providers`, keyed by the provider of the child's resolved model) > `promptInheritance` default > `full`.** The default changes nothing for anyone who does not opt in; a provider-scoped rule leaves same-API children — where the shared prefix is worth the most — untouched.

Design details worth review:

- **Context files ride in the portable identity** because the child loader runs with `noContextFiles: true` — this is the only way a portable child sees project instructions at all.
- **Skills stay un-inherited** — the child loads its own catalogue, same reasoning as ADR 0006's tail-stripping.
- **An absent or empty capture falls back to the generic base, never to the full prompt** — opting into portable must never silently re-embed the base it exists to avoid.
- **Model resolution moved ahead of prompt assembly** in `assembleSessionConfig`, since the child's provider scopes the rules (per-spawn `model` overrides therefore scope correctly too).

## Why

The issue has the full bisection; short form: Anthropic's subscription OAuth gate classifies requests by system-prompt content, and the co-occurrence of two phrases in pi's documentation-routing line (`custom providers (docs/custom-provider.md)` + `pi packages (docs/packages.md)`) in the forwarded append is scored as a third-party app. A full-inheritance bridge child is refused outright while its parent, on the same provider, token, and Claude Code binary, passes — because the bridge only ever projects the parent's *portable* parts for the parent. Portable inheritance gives the child exactly those parts.

## Docs & tests

- ADR 0008 (proposed) records the reasoning and the not-yet-taken further step (provider-declared policy, which this setting is shaped to override later).
- `docs/configuration.md`: `inherit_prompt` frontmatter row, a *Portable inheritance (opt-in)* subsection with the strategy table, and `promptInheritance` settings docs (both forms, precedence).
- 25 new tests across prompts / parent-snapshot / settings / session-config / runtime / custom-agents; full suite green locally (1581), `tsc` and `biome` clean.

## Verification

Live on pi-claude-bridge, same machine and OAuth token: full-inheritance child → 400 (pre-credit; with credit, extra-usage draw in the rate-limit telemetry); portable child — via frontmatter or the provider rule, using the built-in `general-purpose` agent with no per-agent config — succeeds on plan usage. A `@tintinweb/pi-subagents` cross-check (thin-standalone replace semantics) is recorded in the issue as the ecosystem control.

### DIFF
--- a/packages/pi-subagents/docs/configuration.md
+++ b/packages/pi-subagents/docs/configuration.md
@@ -39,6 +39,20 @@ Inheriting the parent's copies instead would give such a child a catalogue of sk
 
 Inheriting the identity rather than the whole prompt also keeps the child's leading text byte-identical to the parent's, which prefix-caching providers and local inference engines reuse instead of reprocessing.
 
+### Portable inheritance (opt-in)
+
+The identity layers include Pi's own preamble and tool guidelines — correct when the child talks to the same raw API as its parent, wrong when the child's provider **re-homes** the prompt into another harness (for example pi-claude-bridge, which projects the child's prompt onto Claude Code's as an append).
+For those children, `inherit_prompt: portable` (frontmatter) or a `promptInheritance` provider rule (settings) switches the inherited identity to the parent's **portable parts** only: project context files, added guideline bullets, and custom/append prompts.
+Skills remain un-inherited (the child builds its own catalogue), and context files must ride along because the child session never loads its own.
+
+| Strategy   | Child's identity                                    | Shares prefix with parent | Use when                                        |
+| ---------- | --------------------------------------------------- | ------------------------- | ----------------------------------------------- |
+| `full`     | parent identity layers, byte for byte (default)     | yes                       | child talks to the same raw API as the parent   |
+| `portable` | parent's context files + custom/append + guidelines | no                        | child's provider re-homes the prompt elsewhere  |
+
+Resolution: agent frontmatter `inherit_prompt` > the `promptInheritance` provider rule for the child's resolved provider > the `promptInheritance` default.
+The reasoning and the provider-scoping rationale are recorded in [ADR 0008](decisions/0008-portable-prompt-inheritance.md).
+
 If you write extensions that add to the system prompt, see [Extensions that append to the system prompt](../README.md#extensions-that-append-to-the-system-prompt).
 The reasoning behind the boundary is recorded in [ADR 0006](decisions/0006-inherited-prompt-is-identity-only.md).
 
@@ -90,19 +104,20 @@ subagent({ subagent_type: "auditor", prompt: "Review the auth module", descripti
 
 All fields are optional — sensible defaults for everything.
 
-| Field               | Default        | Description                                                                                                                                                                                                                                                                                                             |
-| ------------------- | -------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `description`       | filename       | Agent description shown in tool listings                                                                                                                                                                                                                                                                                |
-| `display_name`      | —              | Display name for UI (e.g. widget, agent list)                                                                                                                                                                                                                                                                           |
-| `tools`             | all 7          | The agent's complete tool allowlist — built-in or extension-registered names. `none` for no tools. See [Tool selection](#tool-selection)                                                                                                                                                                                |
-| `model`             | inherit parent | Model — `provider/modelId` or fuzzy name (`"haiku"`, `"sonnet"`)                                                                                                                                                                                                                                                        |
-| `thinking`          | inherit        | off, minimal, low, medium, high, xhigh, max. An unrecognized value is dropped, and the agent inherits the parent's level                                                                                                                                                                                                |
-| `max_turns`         | unlimited      | Max agentic turns before graceful shutdown. `0` or omit for unlimited                                                                                                                                                                                                                                                   |
-| `prompt_mode`       | `append`       | `replace`: parent prompt is the cacheable base; body is appended last with full control (no `<sub_agent_context>` bridge, no `<agent_instructions>` wrapper). `append`: parent prompt is the base; body is wrapped in `<agent_instructions>` and a sub-agent context bridge is injected (agent acts as a "parent twin") |
-| `inherit_context`   | `false`        | Fork parent conversation into agent                                                                                                                                                                                                                                                                                     |
-| `run_in_background` | `false`        | Run in background by default                                                                                                                                                                                                                                                                                            |
-| `enabled`           | `true`         | Set to `false` to disable an agent (useful for hiding a default agent per-project)                                                                                                                                                                                                                                      |
-| `locked`            | —              | Fields a `subagent` tool caller may not override. `true` or a list of field names. See [Locking fields against callers](#locking-fields-against-callers)                                                                                                                                                                |
+| Field               | Default        | Description                                                                                                                                                                                                                                                                                                                                                                                                                        |
+| ------------------- | -------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `description`       | filename       | Agent description shown in tool listings                                                                                                                                                                                                                                                                                                                                                                                           |
+| `display_name`      | —              | Display name for UI (e.g. widget, agent list)                                                                                                                                                                                                                                                                                                                                                                                      |
+| `tools`             | all 7          | The agent's complete tool allowlist — built-in or extension-registered names. `none` for no tools. See [Tool selection](#tool-selection)                                                                                                                                                                                                                                                                                           |
+| `model`             | inherit parent | Model — `provider/modelId` or fuzzy name (`"haiku"`, `"sonnet"`)                                                                                                                                                                                                                                                                                                                                                                   |
+| `thinking`          | inherit        | off, minimal, low, medium, high, xhigh, max. An unrecognized value is dropped, and the agent inherits the parent's level                                                                                                                                                                                                                                                                                                           |
+| `max_turns`         | unlimited      | Max agentic turns before graceful shutdown. `0` or omit for unlimited                                                                                                                                                                                                                                                                                                                                                              |
+| `prompt_mode`       | `append`       | `replace`: parent prompt is the cacheable base; body is appended last with full control (no `<sub_agent_context>` bridge, no `<agent_instructions>` wrapper). `append`: parent prompt is the base; body is wrapped in `<agent_instructions>` and a sub-agent context bridge is injected (agent acts as a "parent twin")                                                                                                            |
+| `inherit_prompt`    | `full`         | What the child inherits as its prompt identity. `full`: the parent's assembled prompt (minus pi's per-session layers) — byte-stable cache prefix with the parent. `portable`: only the parent's portable parts (context files, custom/append prompts, added guidelines) — for children whose provider re-homes the prompt into another harness. The `promptInheritance` setting supplies the default for agents that omit this key |
+| `inherit_context`   | `false`        | Fork parent conversation into agent                                                                                                                                                                                                                                                                                                                                                                                                |
+| `run_in_background` | `false`        | Run in background by default                                                                                                                                                                                                                                                                                                                                                                                                       |
+| `enabled`           | `true`         | Set to `false` to disable an agent (useful for hiding a default agent per-project)                                                                                                                                                                                                                                                                                                                                                 |
+| `locked`            | —              | Fields a `subagent` tool caller may not override. `true` or a list of field names. See [Locking fields against callers](#locking-fields-against-callers)                                                                                                                                                                                                                                                                           |
 
 The caller decides, and the agent file fills the gaps.
 A `subagent` tool parameter wins over the agent file's value for `model`, `thinking`, `max_turns`, `inherit_context`, and `run_in_background`; the agent file supplies whichever of those the caller left unset.
@@ -201,7 +216,24 @@ Two files, merged on load:
   Written by `/subagents:settings`.
 
 **Precedence:** project overrides global on any field present in both.
-Missing fields fall back to the hardcoded defaults (max concurrency `4`, default max turns unlimited, grace turns `5`, consumed-session retention `10` minutes, unconsumed-session retention `720` minutes, abort-all-on-interrupt `true`, mid-run updates `true`).
+Missing fields fall back to the hardcoded defaults (max concurrency `4`, default max turns unlimited, grace turns `5`, consumed-session retention `10` minutes, unconsumed-session retention `720` minutes, abort-all-on-interrupt `true`, mid-run updates `true`, prompt inheritance `full`).
+
+**Prompt inheritance** — `promptInheritance` selects what children inherit as their prompt identity (see [Portable inheritance](#portable-inheritance-opt-in)).
+Either form works:
+
+```json
+"promptInheritance": "portable"
+```
+
+```json
+"promptInheritance": {
+  "default": "full",
+  "providers": { "claude-bridge": "portable" }
+}
+```
+
+The `providers` map keys on the provider id of the **child's resolved model** — only children whose requests actually go through that provider change strategy, so same-API children keep the byte-identical cache prefix.
+An agent's `inherit_prompt` frontmatter wins over both forms.
 
 **Example — global defaults for a beefy machine:**
 
--- a/packages/pi-subagents/docs/decisions/0008-portable-prompt-inheritance.md
+++ b/packages/pi-subagents/docs/decisions/0008-portable-prompt-inheritance.md
@@ -1,0 +1,55 @@
+---
+status: proposed
+date: 2026-09-05
+---
+
+# 0008 — Portable prompt inheritance for children whose provider re-homes the prompt
+
+## Status
+
+Proposed.
+Extends [ADR 0006](0006-inherited-prompt-is-identity-only.md) with a second, opt-in inheritance strategy; changes no default behavior.
+
+## Context
+
+ADR 0006 settled *which parts* of the parent's prompt a child inherits: the identity layers, byte for byte, so a child shares a prefix with its parent that prefix-caching providers reuse.
+The identity layers include Pi's own preamble and tool guidelines.
+
+That is the right trade when the child talks to the same raw API as its parent.
+It is the wrong trade when the child's provider *re-homes* the prompt — forwards Pi's system prompt into another harness that composes it onto its own. pi-claude-bridge does exactly this: it projects the child's Pi prompt into Claude Code's `--append-system-prompt`, where Pi's preamble rides as an append on top of Claude Code's preset.
+
+Re-homing turned out to be more than cosmetic.
+Anthropic's subscription OAuth gate classifies requests by system-prompt content, and flags the combination of two phrases in Pi's documentation-routing line (`custom providers (docs/custom-provider.md)` alongside `pi packages (docs/packages.md)`) as a third-party app.
+A child whose inherited identity reaches that gate is rejected outright while the account has no extra-usage credit — the parent and every non-inheriting path on the same provider, token, and Claude Code binary pass.
+The parent passes for a structural reason: pi-claude-bridge already projects only the *portable* parts of the parent's prompt (context files, skills, custom/append), never Pi's base.
+
+Two facts follow.
+
+1. The information a child needs from the parent's prompt — project context, persona, added guidelines — is exactly the portable parts.
+   Pi's preamble and tool guidelines are the harness's own instructions; a re-homing host supplies its own equivalents, so the inherited copy is at best duplicated tokens.
+2. Pi already reports the portable parts authoritatively: `before_agent_start` carries `systemPromptOptions` (`contextFiles`, `customPrompt`, `appendSystemPrompt`, `promptGuidelines`).
+   No prompt-text heuristics are required to separate them from the base.
+
+## Decision
+
+Add a second inheritance strategy, `portable`, selected per child:
+
+- **Agent frontmatter** — `inherit_prompt: portable | full` wins over everything.
+- **Provider rule** — the `promptInheritance.providers` setting maps the *child's resolved model's provider id* to a strategy, because re-homing is a property of the provider the child's requests actually go to.
+- **Default** — `promptInheritance` (simple string form or `rules.default`), ultimately `full`.
+
+A portable child's identity is built from the parent's `before_agent_start.systemPromptOptions`: context files (the child loader runs with `noContextFiles: true`, so this is the only way they reach the child at all), `promptGuidelines`, `customPrompt`, `appendSystemPrompt`.
+Skills stay un-inherited — the child rebuilds its own catalogue, as in ADR 0006.
+An absent or empty capture falls back to the generic base rather than to the full prompt: opting into portable must never silently re-embed the base it exists to avoid.
+
+The default remains `full`.
+Nobody's children change behavior unless a frontmatter key or a settings entry opts them in, and a provider-scoped rule leaves every other provider's children — including same-API children, where the shared prefix is worth the most — untouched.
+
+## Consequences
+
+- A portable child's prompt shares no byte prefix with its parent.
+  For a re-homed child that prefix was never reusable — the request leaves through a different harness — so nothing is lost; for a same-API child the rule should simply not select portable.
+- Context files reach portable children through the inherited identity instead of being suppressed-and-absent, which also makes them available to `prompt_mode: replace` children that opt in.
+- The resolution happens in `assembleSessionConfig` after the child's model is settled, so a per-spawn `model` override is scoped correctly by the provider it actually selects.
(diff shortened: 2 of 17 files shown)


gotgenes:
Thanks for this, @georgeharker — the bisection in #883 is unusually thorough, and the PR is the cleanest third-party submission this repo has had. `pnpm run check`, `pnpm run lint`, and the full `pi-subagents` suite (1581 tests) all pass on `f5d4ba9d` in a clean worktree.

I confirmed the defect on current `main` before looking at the diff. A throwaway test built a parent prompt with pi's own `buildSystemPrompt` from the pinned SDK `0.84.4` and pushed it through `buildAgentPrompt`: the documentation-routing line survives into the child verbatim in both `append` and `replace` mode, along with `operating inside pi, a coding agent harness`. That is `inheritedIdentity` working as ADR 0006 designed it — the cut is at the skills catalogue or the cwd footer, and pi's base preamble is on the kept side by construction.

Before deciding, I traced the failure into `pi-claude-bridge`, and I think the mechanism is narrower than either of us had it.

The bridge already ships the projection fix. `diag/EXTRA-USAGE-400.md` describes treating prompt captures as an inheritance graph, and `findInheritedPrompts` is present in the published 0.7.0 tarball — the version you reported. So the capability you need is there.

What defeats it is a byte mismatch, not instance isolation. The matcher is an exact substring search for the parent's *full* assembled prompt:

```ts
for (let start = custom.indexOf(key); start !== -1; start = custom.indexOf(key, start + key.length))
```

`key` carries the skills catalogue and the `Current working directory:` footer. `inheritedIdentity` truncates both before embedding, so `indexOf` returns `-1`, no edge is recorded, and the whole base is forwarded. The dates fit: `449078d0` (strip cwd footer) landed 2026-07-24, 0.7.0 shipped 2026-08-08, and `610a4e9a`/`49f3e46b` (strip skills catalogue) landed 2026-08-30. The write-up also describes the child prompt as carrying "memory, and preloaded skills", which is `tintinweb/pi-subagents` — we removed memory in `6ebeb91f` and skill-preloading in `93266ff4`.

On the per-session-instance point in #883: `promptCaptures` is a module-level singleton in the bridge's `src/index.ts`, and pi's `loadExtensionsCached` reuses the cached factory for the same cwd, so parent and child share the map in the ordinary case.

I'd like to hear from you and from @elidickinson before landing anything, because a relaxed matcher on the bridge side, or exposing provenance on ours, may be a smaller fix than a second inheritance strategy.

One constraint on the provenance route, since you raised it as suggestion 2: any marker placed before the inherited text breaks the byte-identical parent prefix that `f35e7b1b` (#180) and `1cc25cf0` (#400) exist to preserve. That's a real tension, and I suspect it's why you reached for a side channel.

Holding this PR open rather than closing it. If we implement on either path, your commits carry `Co-authored-by`. Notes I'd want to fold in either way: `ctx.getSystemPromptOptions()` is command-context-only, so your `before_agent_start` capture is the only seam available — I checked `createContext()` in pi's `runner.ts` and the pinned `0.84.4` declarations. And a `portable` child loses `Available tools:` and the computed tool guidelines entirely, because `systemPromptOverride` becomes pi's `customPrompt` and replaces the base; that matters more than it looks, since the bridge blocks Claude Code's own tools and serves pi's over MCP.

georgeharker:
Oh that is interesting! Let me experiment with that approach too. I'd love to hear from @elidickinson also. I'll raise an issue on pi-claude-bridge to loop into this.

georgeharker:
Verified your trace on every checkable point, and prototyped the bridge-side fix — I think you're right that it's the smaller one.

**Verification first.** My captured query agrees with your byte-mismatch mechanism: zero occurrences of the cwd footer in the child's forwarded append, no catalogue inside the embedded region, and `promptCaptures` sharing confirmed the maps were common to parent and child in that run — bytes, not isolation. Correction posted on #883 (instance isolation survives only where a fresh module instance actually loads — see the scope note below).

**A reconciliation worth having on the record:** my own bridge fork's history documents the pre-#812 world. `fix/subagent-provider-registration` (`ff505ac`, Aug 25) switched child streaming to the child's *own* module instance precisely because routing through the parent's broke capture — its commit message names your `"prompt-capture: no capture for this system prompt"` throw from the other side. That was the bridge-side resolution of #812 I'd proposed; your `f805ffc` resolved it pi-subagents-side instead, children now use the parent's registered provider object, the singleton is shared — and bytes became the blocker. Both analyses are correct for their respective configurations, which is what made this chase itself so confusing.

**The prototype** — [elidickinson/pi-claude-bridge#88](https://github.com/elidickinson/pi-claude-bridge/issues/88), branch [`fix/tail-stripped-inheritance-key`](https://github.com/georgeharker/pi-claude-bridge/tree/fix/tail-stripped-inheritance-key) (`83621a3`): `record()` derives a `tailStrippedPrompt` — the assembled prompt cut at the same boundaries `inheritedIdentity` cuts at — and `findInheritedPrompts` matches under either key, both exact substring searches. Live result against **upstream pi-subagents, default full inheritance**: the child's forwarded append drops from ~6.2k to ~0.9k chars, pi's base and both trigger phrases gone, the parent's portable parts substituted in place. One file plus one plumbing line, no modes, no protocol — and the child's pi-side prompt is untouched, so the byte-identical prefix your #180/#400 rationale protects is preserved. Issue filed with @elidickinson looping him in per your ask.

**Scope:** that covers the shared-capture path. Worktree children (cwd change clears pi's extension cache → fresh module instance, empty captures) still need the provenance side channel — which is where your constraint analysis and my `before_agent_start` capture converge, so option 2 stays live for that case regardless.

**On this PR:** I'll fix the `Available tools:` gap you flagged regardless of outcome — `assembleSessionConfig` has the child's `toolNames` at assembly time, so a portable identity can carry its own tool section instead of the parent's stale one. But if the bridge-side fix lands, portable stops being *necessary* for #883 and becomes what it arguably should be: an opt-in for re-homing hosts that don't (or can't) fix their projection. Happy to reframe the PR that way, or withdraw it in favor of the bridge fix plus a provenance follow-up — your call, and @elidickinson's.

gotgenes:
Thanks, @georgeharker — the correction on #883, the reconciliation, and a prototype filed upstream inside an hour is more than I asked for.

The `ff505ac` / `f805ffc` reconciliation is the part I'm glad is on the record. I had read your instance-isolation claim as simply wrong; it was true of your fork's configuration and false of ours, and neither of us could see the other's. That is worth remembering the next time a cross-repo diagnosis looks contradictory.

Two things to add, both usable whichever way this lands.

**The worktree path fails silently, not loudly.** I'd assumed an empty capture map would surface as your `no capture for this system prompt` throw. It won't: the child's own `before_agent_start` records a self-capture, so the resolver finds an exact match and simply has no inherited edge. The base is forwarded verbatim and the symptom is the same silent 400, which makes that residual harder to notice than the error path would be. I confirmed `useExtensionCacheCwd` clears the cache on any cwd change, so the mechanism you describe is right.

**The `Available tools:` reconstruction needs snippets, not names.** `registry.getToolNamesForType(type)` gives you names only, and pi renders `- ${name}: ${toolSnippets[name]}` — so rebuilding from `toolNames` would produce a bare list where pi produces a described one.

The snippets are already in the capture you take. `BuildSystemPromptOptions.toolSnippets?: Record<string, string>` is populated in `_baseSystemPromptOptions` beside `selectedTools`, so `before_agent_start` carries them. Filtering the parent's snippets by the child's `toolNames` reproduces pi's list exactly, because pi itself only lists tools that have a snippet:

```ts
const visibleTools = tools.filter((name) => !!toolSnippets?.[name]);
```

The only gap is child-only tools — `ask_parent` and `notify_parent` — which the parent never had a snippet for.

**On disposition:** I'd like to leave that to @elidickinson's read of #88, since your table is right that the stripped-key option makes the existing design do what it was built to do, and that is a smaller change than a second inheritance strategy in this package. If it lands, `portable` stops being necessary for #883 and becomes an escape hatch for re-homing hosts that can't fix their projection, plus the worktree case — a narrower warrant, but not zero. If it doesn't, your PR is the fallback and I'd take it with the simplifications from my last comment.

Either way I'm keeping #883 open and cross-linked to #88. The bridge fix mitigates the symptom for bridge users; it doesn't change that a pi-subagents child carries pi's base prompt into any provider that re-homes it, which is our side of the ledger regardless of who fixes it first.

No rush on the PR from my end — it's not blocking anything here.

georgeharker:
**Worktree follow-up: measured end-to-end — your "fails silently" instinct was right, the mechanism runs through `pi-permission-system`, and there's a second bug on this side of the ledger that makes it invisible.**

Everything below is on stock components — no patched pi, no forked pi-subagents; the only instrument is a ~50-line pure-observer extension registered last, logging one JSON line per `before_agent_start` (agent name, final prompt length, the original `systemPromptOptions.customPrompt` length, both `Available tools:` lists, and whether the original override survives verbatim in the final prompt).

## The one-variable A/B

Same worktree-agent spawn, same everything, with and without `npm:@gotgenes/pi-permission-system` in the packages list:

| | WITH permission-system | WITHOUT |
| --- | --- | --- |
| Probe (child turn) | override `survives=false`; tool section **swapped** to the child's allowed set | `survives=true`; tool section **inherited** from the parent intact |
| Bridge (stock debug log) | reentrant child query → `no match for 22157-char system prompt … diverges at offset 412` → **throws** | reentrant child query → `fresh query … tools=8` → **stripped-key match fires** |
| Subagent result | `Agent completed in 2.6s (0 tool uses)` — **empty** | `Agent completed in 14.0s (1 tool use)` — real work, child's own CC session |

The divergence offset is the tell: 412 chars in is the `Available tools:` list. `AgentPrep`'s `sanitizeAvailableToolsSection` replaces the inherited tool list with the agent's allowed set and prunes the tool-dependent guidelines — which is exactly where the child's prompt stops containing the parent's prompt, byte-for-byte.

## Two things worth separating

**1. The rewrite ↔ the byte-stable prefix — it seems like there might be unintended interplay between two of your designs.** The rewrite is doing something *correct*: without it, the child's prompt advertises the parent's 28 tools while its active surface is 8 — inherited tool docs lying to the child, which is the same gap you flagged on this PR ("a portable child loses `Available tools:` entirely"). But the mechanism — post-hoc string surgery on the assembled prompt — defeats ADR 0006's byte-identical parent prefix (#180/#400) for every agent whose active tools differ from the parent's, and it breaks any consumer that matches prompts structurally. The bridge's inheritance matching is one such consumer; the KV-cache prefix rationale is another. I don't think either design is wrong — they are conflating with the bridge requirements.

The convergence worth noting: your own suggestion upthread — rebuild the tools section from the parent's `toolSnippets` filtered by the child's `toolNames`, which `before_agent_start` already carries — is exactly the build-time mechanism that would let the permission system contribute the corrected tool surface without post-hoc surgery, and it would close the portable-mode tool-docs gap on this PR at the same time. The one constraint that makes it fully coherent: **the corrected section must be appended *after* the inherited identity region, not edited within it.** The parent's bytes then stay verbatim at the front — both the KV-cache prefix and structural matchers keep working — the child's honest tool docs follow them, and the projection detail falls out: the bridge's substitution replaces only the identity region, so the child-specific tools section survives into the forwarded append.

**2. Silent failure issue discovered during investigation — a bug independent of the bridge.** A child whose provider throws on its first turn is reported as `Agent completed in 2.6s (0 tool uses)` with an **empty result**, not as a failure. Every full-stack worktree run in our testing failed exactly this way and was masked by it — and because the shell reads like success, the parent models *confabulate*: we collected "replied ok", "created proof.txt with done", "changes are on its worktree branch" — while no branch and no file ever existed (`git branch` clean, working tree clean).

## Corrections to the thread's record, including mine

- Your predicted mechanism — self-capture in a fresh instance, silent verbatim forward — turned out not to be the path. The child's own instance never serves its query at all: the bridge's child instance logs `skipping re-registration, parent instance active` and the query runs reentrant on the parent's instance. And the failure is a **loud** capture throw that gets swallowed downstream by item 2 — so nothing is ever forwarded, no 3p classification occurs on this path. Silent in effect, but via empty-completion, not base-forwarding.
- My earlier scoping was also wrong, and it's on the record — my reply upthread, the analysis on elidickinson/pi-claude-bridge#88, and the #89 PR description all claimed worktree children need the provenance side channel because the fresh instance's capture map is unreachable. Measured, that's not so — the child's assembled prompt embeds the stripped parent key inside the override (the `customPrompt` region), so the parent-instance match bridges the instance gap fine. **The permission-system rewrite is the only match-breaker.** With it removed, bridge PR elidickinson/pi-claude-bridge#89 covers worktree children cleanly — verified with real work done.
- The chain, for completeness (stderr probe on `emitBeforeAgentStart`, extensions in registration order): child base assembled correctly (`customPrompt=7556`, identity intact) → combiner appends → cribsheet appends → svg-mcp appends → **permission-system rewrites 24337 → 21892** → pi-ask appends → 22157 is what the bridge receives.

## Where this leaves things

- **#89** (bridge, stripped-key matching): scope now provably includes worktrees; no change needed there.
- **This PR**: unchanged — portable remains the fallback for re-homing hosts that can't fix their projection, and the tool-docs gap you flagged is still worth fixing here if you'd take it.
- **The rewrite interplay and the silent failure issue**: both this repo's ecosystem, both measured — happy to file the silent-failure one as its own issue if useful. The probe and repro are inline below, so the A/B runs as one command on your side.

---

## Repro (stock pi, stock packages; the probe is the only addition)

**The probe** — pure observer, returns nothing, safe under any stack:

```typescript
// /tmp/prompt-probe.ts — registered LAST so it sees the final prompt.
import { appendFileSync } from "node:fs";

const LOG = "/tmp/prompt-probe.log";

function toolNames(section: string): string {
	const match = section.match(/Available tools:\n((?:- .+\n)+)/);
	if (!match) return "(none)";
	return match[1]
		.trim()
		.split("\n")
		.map((line) => line.slice(2, line.indexOf(":", 2)).trim())
		.join(",");
}

export default function promptProbe(pi: any): void {
	pi.on("before_agent_start", (event: any) => {
		try {
			const sp = event.systemPrompt ?? "";
			const cp = event.systemPromptOptions?.customPrompt ?? "";
			appendFileSync(
				LOG,
				JSON.stringify({
					ts: new Date().toISOString(),
					agent: (sp.match(/<active_agent name="([^"]+)"\/>/) || [])[1] ?? "(parent)",
					finalLen: sp.length,
					customLen: cp.length,
					finalTools: toolNames(sp),
					customTools: cp ? toolNames(cp) : "(no override)",
					overrideSurvives: cp.length > 0 && sp.includes(cp),
				}) + "\n",
			);
		} catch {
			/* never break a turn for a probe */
		}
	});
}
```

**Steps** (the probe must be the *last* entry in the **global** `packages` list — a project-level `.pi/settings.json` entry resolves before global packages, so it would observe the pre-rewrite prompt):

```bash
# 1) append "/tmp/prompt-probe.ts" to the END of packages in settings.json

# 2) worktree-agent spawn (worktreeAgents: ["worktree-agent"] in subagents-worktrees.json):
CLAUDE_BRIDGE_DEBUG=1 pi --model claude-bridge/claude-opus-5 -p 'Use the subagent tool \
  with model "claude-bridge/claude-opus-5", subagent_type "worktree-agent", \
  prompt: Create a file named proof.txt containing the word done. Then reply with just the word "ok".'

cat /tmp/prompt-probe.log          # child row: overrideSurvives=false, tools swapped
grep "no match" ~/.pi/agent/claude-bridge.log   # reentrant child query, diverges at 412
git branch                          # empty — no real work

# 3) remove npm:@gotgenes/pi-permission-system from packages, rerun 2):

cat /tmp/prompt-probe.log          # child row: overrideSurvives=true, tools inherited
grep "no match" ~/.pi/agent/claude-bridge.log   # nothing — the stripped-key match fires
git branch                          # the child did real work this time
```

gotgenes:
Landed as #883 in dba48999a8f6eaf7b76a3d471e9b7ddebe1eb35a, reimplemented rather than merged — the review rounds on this PR narrowed the design in three ways, all recorded in the triage note and the plan:

1. **Provider-keyed setting instead of `inherit_prompt` frontmatter.** A per-spawn `model` override moves a child to a different transport, and frontmatter would keep the wrong strategy across that move. The setting is keyed on the child's resolved provider instead, so the strategy follows automatically. No global `default` arm either — an explicit per-provider map can't accidentally cost a local-inference user (#180's constituency) the shared prefix #890 restored.
2. **`promptGuidelines` dropped from the portable identity.** Pi derives it per session from the tools actually in the registry, so inheriting the parent's would assert guidance for tools the child may not hold — the same defect ADR 0008 removed from `<sub_agent_context>`.
3. **Rendered from parts in Pi's own composition order**, byte-exact against `<project_context>`, rather than assembled with a different order and without Pi's lead-in sentence.

The mechanism you built — the `before_agent_start` capture, the fail-safe fallback to a generic base rather than the full prompt, context files riding along because the child loader suppresses them — is intact and credited throughout: every implementation and docs commit carries `Co-authored-by: George Harker <george@george-graphics.co.uk>`, and ADR 0009 credits the `pi-claude-bridge` bisection methodology too.

Thanks for the thoroughness across every round of this — the bisection, the correction on the byte-mismatch trace, the bridge-side prototype, and the worktree measurement all made it into the final design or its record. Closing this in favor of the reimplementation; #883 has the full writeup.

PR #908, opened 2026-09-10
fix(pi-permission-system): support OMP prompt arrays

### LINKED ISSUE #860: pi-permission-system: error when using with omp
### Package

pi-permission-system

### Package version

29.1.0

### What happened?

When using with omp I get the following error

```
Extension "~/.omp/plugins/node_modules/@gotgenes/pi-permission-system/src/index.ts" error: (prompt || "").replace is not a function. (In '(prompt ||
 "").replace(/\r\n/g, `
 `)', '(prompt || "").replace' is undefined)
```

### Expected behavior

No errors

### Steps to reproduce

1. Run `omp install @gotgenes/pi-permission-system`.
2. Start omp and send a prompt

### Additional context

I have `pi` installed as well and there I have permissions installed and configured as well. Not sure if that's important.

JasonLandbridge:
## Summary

- normalize string and string-array `before_agent_start.systemPrompt` payloads before tool-surface rendering
- pass the normalized prompt through agent-name resolution, skill filtering, and replacement comparison
- add regression coverage for OMP array prompt fragments at the renderer and handler seams

Fixes #860.

## Verification

- `pnpm --filter @gotgenes/pi-permission-system test -- test/exposure/tool-surface-prompt.test.ts test/handlers/before-agent-start.test.ts`
- `pnpm --filter @gotgenes/pi-permission-system run lint`


gotgenes:
Thanks for the update here, and apologies for the wait.

The array-normalization fix is one I want. But this branch now carries a second commit, 9e7a95ae ("add support for extracting target path from Oh My Pi hashline…"), which is the change from #922 — the same two files. I'm not taking that one: parsing another host's bespoke edit-envelope format is a different thing from normalizing a payload shape that violates a contract the package already reads, and I want that boundary written into the package's non-goals before I answer #922 properly.

Could you drop 9e7a95ae and leave #908 as the `systemPrompt` normalization alone? With that, this is a straightforward review.

I've approved the CI run on the current head, so you'll get a result either way.

gotgenes:
One follow-up that should save you a round trip: the red check on this branch isn't the array fix. It's a Biome format nit inside `9e7a95ae` — the commit I asked about above:

```text
packages/pi-permission-system/src/access-intent/tool-input-path.ts format ━━━━━━━━━━
  × Formatter would have printed the following content:
     18    │ - function·getHashlineEditPath(
     19    │ - ··input:·Record<string,·unknown>,
     20    │ - ):·string·|·null·{
        18 │ + function·getHashlineEditPath(input:·Record<string,·unknown>):·string·|·null·{
```

So dropping `9e7a95ae` should clear the check at the same time, and `d0673193` alone ought to go green.

I've closed #922 with the reasoning, and the non-goal it rests on is now in the README and architecture doc ([b9d645c](https://github.com/gotgenes/pi-packages/commit/b9d645cdf39df10afe5493ca1817df3a1c99c527)). The short version: `systemPrompt` coming through as a `string[]` violates a contract this package already reads, so normalizing it is worth doing on its own merits — this half I want.

Issue #909, opened 2026-09-10
pi-permission-system: honor explicit parent forwarding for subprocess children with their own UI


boadij:
### Package

pi-permission-system

### What do you want to change?

An out-of-process subagent that explicitly sets `PI_SUBAGENT_PARENT_SESSION=<parent-session-id>` still handles `ask` permissions in its own Pi TUI when that child has UI.

This means the declared forwarding target is ignored for subprocess agents that intentionally remain visible and interactive for inspection.

I observed this while adding `pi-permission-system` interoperability to [Pi Herdsman](https://github.com/boadij/pi-herdsman), which runs managed agents as separate Pi processes in visible Herdr panes.

The child was correctly identified by `<active_agent name="permission-test-child"/>`, its project-local `permission:` frontmatter was applied, and `git status` correctly resolved to `ask`.

However, the permission dialog appeared in the child pane instead of the lead session named by `PI_SUBAGENT_PARENT_SESSION`.

Current `selectAuthorizer()` behavior explains the result: `ctx.hasUI` is checked before subagent detection, so a subagent with its own UI selects `LocalUserAuthorizer` and adjudicates locally.

The current subagent integration docs describe `PI_SUBAGENT_PARENT_SESSION` as identifying the session the child forwards asks to, while the permission-forwarding section later limits forwarding to children without direct UI access.

That leaves visible subprocess subagents without a way to declare that permission authority should remain in the parent session.

### Why?

Some subprocess-based agent managers deliberately give every child its own visible Pi UI while keeping permission authority centralized in the lead session.

[Pi Herdsman](https://github.com/boadij/pi-herdsman) is one such case: the child pane exists for observability and direct inspection, not to make that child the human approval authority.

For a topology such as `lead -> parent agent -> child agent`, Herdsman preserves direct logical ownership separately while propagating the root lead session as `PI_SUBAGENT_PARENT_SESSION`.

A permission `ask` should therefore surface where the human is already working, rather than blocking a nested child pane and requiring the user to discover and answer the prompt there.

This also makes the out-of-process adapter announcement less useful than it appears: the spawner can provide the documented parent-session target correctly, but the child cannot opt into using it when Pi reports `hasUI`.

### How? (optional)

Please decide the intended authority semantics when an out-of-process child has both its own UI and an explicit `PI_SUBAGENT_PARENT_SESSION`.

[...]

Issue #907, opened 2026-09-10
pi-permission-system：Root session is detected as a subagent when `PI_SUBAGENT_PARENT_SESSION` names its own id — serving heartbeat withdrawn, every forwarded ask fails closed (nicobailon/pi-subagents interop)


gaop154:
### Package

pi-permission-system

### Package

`@gotgenes/pi-permission-system` 31.1.2 (the relevant logic is unchanged in current main per source inspection)

### Environment

- Windows 11, interactive pi TUI session
- `pi-subagents` (nicobailon) 0.66.0 — spawns runner children as separate `pi` processes
- Any `ask`-bearing policy (e.g. gated `write`), child attempts a gated tool call

### What happened

Every `ask` raised inside an out-of-process child is fail-closed with:

```
[pi-permission-system] This 'external_directory_write' call for tool 'write' for agent 'scout'
requires approval, but no interactive UI is available. Reason: Session '01a086cf-…' is not
serving forwarded permission requests.
```

…even though the root interactive session is alive with a UI the whole time.

Review-log timeline for one session:

```
15:36:03  forwarded_permission.serving_started     sessionId=01a086cf…        (session start)
15:36:25  forwarded_permission.serving_stopped     sessionId=01a086cf…        (~22s later, first turn event)
15:37:06  forwarded_permission.request_created     targetSessionId=01a086cf…  (child resolves the parent correctly)
15:37:08  forwarded_permission.no_serving_session  servingChannel=heartbeat servingState=absent
15:37:08  permission_request.denied                decidedBy={"kind":"unavailable"}
```

The parent serves for only ~22 seconds — until the first `before_agent_start` / `input` / `tool_call` after extension load — then stops. Children asking later find `servingState: absent` and give up after the grace window.

### Root cause

1. `nicobailon/pi-subagents` sets `process.env.PI_SUBAGENT_PARENT_SESSION = <root session id>` **inside the root interactive process** at session_start, so its runner subprocesses inherit the parent reference (verified in 0.66.0, `src/extension/index.ts` session_start handler; confirmed live — `printenv PI_SUBAGENT_PARENT_SESSION` in the root session prints the root's own session id).

2. `SUBAGENT_ENV_HINT_KEYS` includes `SUBAGENT_PARENT_SESSION_ENV_CANDIDATES` — "A process that names a parent session is a child by definition" (`src/authority/permission-forwarding.ts`).

[...]

PR #911, opened 2026-09-11
fix(pi-permission-system): preserve serving for UI self-parent hints

mevatron:
## Summary

Addresses the root self-parent-marker misclassification described in https://github.com/gotgenes/pi-packages/issues/907.

- Ignore an inherited parent-session hint only when the context has a UI and the hint matches its valid, normalized current session ID.
- Preserve registry precedence, explicit child markers, distinct-parent hints, filesystem evidence, and headless-child detection (including headless self-reference).
- Leave permission evaluation, forwarding-target resolution, and serving lifecycle logic unchanged.
- Add detection and composition-root regressions and isolate detection environment hints in five affected test fixtures.

The self-reference proposal and the headless-child constraint came from @gaop154's report; the commit includes co-author credit.

## Automated local testing

- Reproduced the original serving withdrawal with the isolated installed-code reproduction before implementing the fix.
- Added a composition-root regression: start the interactive root, set its own parent-session marker, invoke `before_agent_start`, and assert that its serving registration remains present.
- Added detection coverage for both parent-session keys, whitespace normalization, headless self-reference, distinct parents, mixed hints, invalid/unavailable current IDs, registry precedence, and filesystem evidence.
- Reproduced environment-dependent test failures: 22 failures in the interactive parent test context and 24 in a delegated context; clearing inherited detection hints in the test subprocess made the affected 158 tests pass.
- Fixed the fixtures to stub every authoritative `SUBAGENT_ENV_HINT_KEYS` entry before each test and restore the original environment afterward; individual cases explicitly set their required hints.
- Full permission-package suite with all detection hints absent: **157 files, 4,139 tests passed**.
- Full permission-package suite with all 13 detection hints deliberately populated: **157 files, 4,139 tests passed**.
- Workspace `pnpm run check`, `pnpm run lint`, `pnpm run test`, and `pnpm fallow dead-code`: **passed** in fresh pre-completion review.
- Independent Standards and Spec reviews found no blocking code issues.

## Local live background-child testing

Inspected the operator's September 11 session `01a08e17-1f67-70a9-b927-03a866899420` and the child transcripts/launch receipts, rather than relying only on agent summaries.

The `permission-e2e` agent used `tools: bash` and explicitly configured the patched Git checkout's `packages/pi-permission-system/src/index.ts` through `subagentOnlyExtensions`.
That checkout resolves to this PR's commit, `11ea9e38eba19b8fda9d2ac083a7ee6ff4d414c1`.
Both launches used `async: true`, fresh context, shared cwd, and no worktree.
Both launch receipts included the configured patched entry-point identity (`sha256:191d32dcad00cc28`) and allowed ambient discovery.

| Probe | Run ID | Actual child tool result |
| --- | --- | --- |
| `printf 'permission-e2e-probe\n'` | `87b8ed4e-17a8-416c-920f-65fc9df3ffe1` | `permission-e2e-probe` followed by a newline; `isError: false` |
| `rm ./ask-to-remove-me` on an operator-created disposable file | `3d723d39-1ee8-458b-880a-60014cac3554` | `(no output)`; `isError: false` |

These runs confirm explicit patched-extension selection and successful command execution by actual background children.
The supplied transcript does **not** record the human dialog choices or correlated forwarding review-log events, and it does not establish that the `printf` probe matched an `ask` rule.
Accordingly, this PR does not claim a fully evidenced live deny-then-approve round trip from those artifacts alone.

Earlier foreground removal probes were not equivalent forwarding validation: their receipts showed only the subagent prompt runtime and no permission extension; one child also substituted Python `unlink()` for `rm`.
Those results are not counted as successful permission tests.

## Scope and remaining limitations

- This is a narrow repair, not a claim that every finding in #907 is resolved.
- Session-ID churn, stale serving/target identities, Windows heartbeat rename failures, and #909 authority routing remain outside this change.
- Child extension discovery/injection is a separate integration concern; this live test explicitly selected the patched extension.
- The new composition test asserts serving-registry retention, not a complete filesystem-heartbeat refresh round trip.
- Pre-completion review returned WARN for unavailable Mermaid rendering tooling (unchanged diagrams), an additional reviewer-only forced-environment probe blocked before execution, and the live approval-evidence limitation above; the implementer's full contaminated-environment suite did pass.

Refs #907

### DIFF
--- a/packages/pi-permission-system/docs/architecture/architecture.md
+++ b/packages/pi-permission-system/docs/architecture/architecture.md
@@ -509,8 +509,9 @@ This requires two detections:
    Each concurrent sibling child of the same parent receives a unique session id from `sessionManager.newSession()`, so siblings occupy distinct keys - one sibling's `disposed` event cannot evict another's entry (fixes #298).
    The registry is a process-global singleton (via `getSubagentSessionRegistry()`, backed by `globalThis` + `Symbol.for()`) because each session's `ResourceLoader` creates its own `pi.events` bus: the parent's instance registers the child over the parent bus, while the child's separate jiti instance reads the same global store to detect itself and resolve its forwarding target.
 2. **Env vars** (`SUBAGENT_ENV_HINT_KEYS`) - returns `true` when any key is set to a non-empty, non-whitespace value.
-   Used by process-based subagent extensions.
-   The list is composed from the per-extension markers plus `SUBAGENT_PARENT_SESSION_ENV_CANDIDATES`, since a process that names a parent session is a child by definition - which is what makes the convention's single out-of-process obligation sufficient on its own (#789).
+   A UI host's parent-session hint is ignored only when it matches that host's own non-empty session id.
+   All other hints remain child evidence, including a matching parent-session hint in a headless process.
+   The list is composed from the per-extension markers plus `SUBAGENT_PARENT_SESSION_ENV_CANDIDATES`, so naming a parent session remains the convention's single sufficient out-of-process obligation (#789).
 3. **Filesystem path** - session-directory path-based fallback (child session dir is nested under `subagentSessionsDir`).
 
 ### Parent-session resolution (`resolvePermissionForwardingTargetSessionId`)
--- a/packages/pi-permission-system/docs/subagent-integration.md
+++ b/packages/pi-permission-system/docs/subagent-integration.md
@@ -51,7 +51,9 @@ PI_SUBAGENT_PARENT_SESSION=<parent-session-id>
 ```
 
 That is the whole obligation.
-The variable identifies the session the child forwards its asks to, and naming a parent session is itself sufficient to mark the process as a child — a separate "I am a subagent" marker is neither required nor expected.
+The variable identifies the session the child forwards its asks to, and naming a parent session is itself sufficient to mark a process as a child — a separate "I am a subagent" marker is neither required nor expected.
+For compatibility with extensions that export the marker from their interactive host, the permission detector ignores it only when it matches that UI host's own session id.
+A headless process receives no such exception.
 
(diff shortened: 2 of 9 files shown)


gotgenes:
Thanks, @mevatron — this correctly diagnosed the root cause: `nicobailon/pi-subagents` sets `PI_SUBAGENT_PARENT_SESSION` to the root's own session id inside the root process, which this repo's `isSubagent` hint list misread as child evidence.

This repo reimplements adopted third-party changes through its own TDD cycle rather than merging them, so the fix landed as three commits on `main` (0e1188bb88b8d5b3ee14db3027baf4038644678b and siblings), two of them carrying `Co-authored-by: mevatron <mevatron@users.noreply.github.com>` for the diagnosis and fix your PR contributed. Your env-hygiene work on the five affected test fixtures shipped too, as its own `test:` commit.

One place the shipped fix diverges from this PR: instead of adding `hasUI` to `SubagentDetectionContext` and comparing the marker against the UI host's own session id at the predicate, the fix moves to the guard's one consumer — `ForwardingManager.start` now serves whenever `ctx.hasUI` is true, full stop, matching the rule `selectAuthorizer` already applies to the ask path.

The reason for preferring that over the equality comparison came from reading `nicobailon/pi-subagents` at `v0.66.0` and `v0.67.0`: it sets the marker only from `resetSessionState`, reached only from the `session_start` handler. After a mid-session session-id change the marker still holds the previous id, so an equality comparison stops matching and the root withdraws serving again. Dropping the predicate consult entirely is immune to that. A separate commit also makes the serving heartbeat re-resolve the live session id each poll tick, which covers the same churn from the other side.

Closing as shipped rather than merged, per that convention. Appreciated the diagnosis, and the thoroughness of the reproduction notes.

mevatron:
@gotgenes Awesome! Thanks for making this plugin, I use it daily :grin: , and glad I could do a small part to help out!

PR #922, opened 2026-09-14
fix(pi-permission-system): parse Oh My Pi edit target paths

JasonLandbridge:
## Problem

Oh My Pi's `edit` tool sends hashline edits as a text envelope instead of the native `input.path` shape.

Example:

```text
[src/index.ts#A1B2]
PUT 1.=1:
+replacement
 ```

 The permission system only inspected input.path, so these requests were treated as pathless edit calls. Path-specific edit rules could not match, the prompt displayed with edit
 input, and session approvals fell back to the generic edit-tool scope.

 Fix

 - Preserve the existing input.path extraction behavior.
 - Parse the target path from the leading [path#hash] header when the native path is absent.
 - Reuse getPathBearingToolPath() from getToolInputPath() so per-tool and cross-cutting path gates use the same extraction logic.
 - Leave permission defaults and native tool behavior unchanged.

 Verification

 - Added a regression test for an Oh My Pi hashline edit envelope.
 - Confirmed the regression test failed before the implementation.
 - Full package suite passes: 161 test files, 4,193 tests.
 - Tested the linked package through OMP; path-aware edit permissions now work correctly.

 Scope

 Changes are limited to:

 - packages/pi-permission-system/src/access-intent/tool-input-path.ts
 - packages/pi-permission-system/test/access-intent/tool-input-path.test.ts


gotgenes:
Thanks for this, and for the detailed problem statement — it did real work even though I'm declining the patch. It's what made the boundary clear to me.

I've now written the non-goal I mentioned on #908, in [b9d645c](https://github.com/gotgenes/pi-packages/commit/b9d645cdf39df10afe5493ca1817df3a1c99c527). The line it draws:

> The host boundary is a line between two kinds of change, not a refusal to accept fixes: a payload that violates a contract this package already reads is normalized defensively, while a foreign host's bespoke tool formats, host-only tools, and separate approval authority are each declined.

`systemPrompt` arriving as a `string[]` is the first kind. Pi's `BeforeAgentStartEvent` declares `systemPrompt: string`, so normalizing an array is defensive handling of input that violates a contract this package already reads — correct whoever sent it. That's #908, and I want it.

The hashline envelope is the second kind. `[src/index.ts#A1B2]` is Oh My Pi's format, and parsing it means this package tracks another host's edit-tool syntax with no lane here that executes against that host to tell me when it changes.

What pushed me off the fence was your own finding:

> Path-specific edit rules could not match... session approvals fell back to the generic edit-tool scope.

That's a silent degradation on a permission system, which is strictly worse than the crash in #860 — a crash announces itself. Partially supporting a host I can't test against produces that class of failure quietly, and I'd rather decline than ship a guarantee I can't execute against.

The boundary is conditional, not permanent. I've written down [the five conditions that would make a second host a goal](https://github.com/gotgenes/pi-packages/blob/main/packages/pi-permission-system/docs/architecture/architecture.md#what-would-make-a-second-host-a-goal); two of them aren't mine to satisfy.

For what it's worth, I think the strongest place for this fix is upstream in OMP's compatibility layer. It already remaps `@earendil-works/pi-*` imports at load time, so adapting shapes there would fix every Pi extension at once rather than one at a time.

Closing this one. Thanks again — and #908 without `9e7a95ae` is a change I'd like to land.

Issue #928, opened 2026-09-16
MCP permission targets: prefix-named tools derive no server; proxy-shaped MCP clients are invisible to the mcp surface


georgeharker:
# Upstream issue draft — file one issue covering both PRs

Title: **MCP permission targets: prefix-named tools derive no server; proxy-shaped MCP clients are invisible to the `mcp` surface**

---

## The problem, in one picture

The `mcp` permission surface can only see MCP calls whose identity it can
reconstruct from the proxy tool's *arguments*. Two gaps make that
reconstruction fail today:

1. **Prefix-named tools derive no server.** The `mcp()` proxy carries `tool`
   values like `chrome_devtools_take_screenshot`, and aggregators
   (mcp-combiner) expose `<server>_<tool>` names — but
   `addDerivedMcpServerTargets` only derives server candidates for
   *suffix*-named tools and explicitly skips prefix-named ones. Exact-server
   rules (`mcp: {"github": "deny"}`) silently never fire.
2. **Only the literal `mcp` tool is routed to the surface at all.**
   `classifyToolKind` checks the name `"mcp"` — any other proxy-shaped MCP
   client (a renamed proxy, an aggregator's gateway tool) is gated as a
   generic extension tool: no per-server rules, no tool-pattern MCP rules, no
   `target` evidence for authorizers. Switching MCP client, or running two,
   silently loses permission granularity — a security regression that looks
   like a naming difference.

## How a proxied MCP call becomes rule targets — step by step

Walkthrough for `mcp({ tool: "github_search_code", args: {...} })`:

1. **The gate sees `(toolName, input)`.** The proxy collapses hundreds of MCP
   tools into one registered tool, so the identity the operator wants to gate
   on (server + tool) lives inside the input JSON, not the tool name.
2. **Surface routing.** `classifyToolKind("mcp")` → the `mcp` surface; every
   other name → the generic extension surface (gap 2). *(PR 2: a registered
   proxy name routes here too.)*
3. **Verb extraction from the input fields** — `tool` → call, `search` /
   `describe` → discovery, `connect` → connection, bare `server` → list,
   empty → status. *(PR 2: the registered descriptor's `describeInvocation`
   does this for its own arg shape; an explicit `server` it supplies
   short-circuits step 4.)*
4. **Server derivation, when the call carries no explicit `server`:**
   1. qualified split — `server:tool` needs no derivation;
   2. **prefix convention** — longest configured server that is the leading
[...]

PR #929, opened 2026-09-16
feat(pi-permission-system): derive the bare server for prefix-named MCP tools; stop re-prefixing

georgeharker:
# PR 1 — derive the bare server for prefix-named MCP tools; stop re-prefixing

> Part of #928 — the step-by-step walkthrough of how proxied MCP calls
> become rule targets, the per-style candidate tables, and the rule-author
> guidance live there. This PR is the derivation half.

## What changes

`addDerivedMcpServerTargets` + `pushMcpToolPermissionTargets` in
`src/access-intent/mcp-targets.ts`:

1. A leading `<server>_` segment now derives the bare server candidate (no
   re-prefixed forms — the name is already qualified). Prefix names are the
   common case in practice: the `mcp()` proxy carries `tool` values like
   `chrome_devtools_take_screenshot`, and mcp-combiner's combined names are
   `<server>_<tool>`. Today exact-server rules (`mcp: {"github": "deny"}`)
   never fire for them unless the call also passes an explicit `server`.
2. Prefix derivation is **longest-match only** (the configured list is ordered
   longest-first): `foo_bar_baz` belongs to `foo_bar`, never also to `foo`.
   A prefix hit also suppresses suffix coincidence — one naming convention
   per name, disambiguated.
3. An explicit `server` on an already-prefixed name no longer produces
   redundant re-prefix candidates (`github_github_search_code`) that match
   nothing and push useful candidates down the ordered list.
4. Suffix-named behavior unchanged.

Docs: `docs/configuration.md` gains the derivation walkthrough and the
which-rule-shape-to-write table under the `mcp` surface section.

## Tests

Six new cases in `test/access-intent/mcp-targets.test.ts`: prefix derivation,
unconfigured-server negative, re-prefix removal, qualified-name preservation,
longest-match-only, and prefix-suppresses-suffix. Full package suite:
4261/4261.


gotgenes:
Thanks again for this, @georgeharker — and for the thorough walkthrough in #928.

I adopted the capability with a simplified design rather than merging this branch directly. What's landed:

- A prefix-named tool derives the bare server as a permission candidate, exactly as this PR does, with the longest-match and prefix-suppresses-suffix rules you proposed: 1994b3227e16898c8ca4ff0d7a17b258dc306bcb
- The redundant re-prefixed candidates your PR also removed: 72b3b0b6859511f9ce41d55aa803e57e5886077c
- Docs describing the derivation and rule-shape guidance: 49484a635d69292e8e0ad17dbe792019acd97269

What changed from the PR: while confirming the fix, I found the deeper cause underneath it — the `mcp` surface's matching stopped at the first candidate name that matched any rule, rather than honoring last-match-wins across the whole candidate list the way this package's README has always documented. That let a catch-all mask a more specific rule regardless of this PR's fix, including in the package's own documented `mcp` config example. c57369648a4ffd64de9b07999e2687ceaa96e867 fixes that first, and the candidate order your tests pinned (`targets[0]` as the bare server) is reversed to tool-name-first, since order now only decides what's reported in the prompt rather than which rule wins.

Closing this in favor of that implementation, credited above. Refs #928, #946.

PR #930, opened 2026-09-16
feat(pi-permission-system): registerMcpProxy — let MCP proxy tools declare their structure

georgeharker:
# PR 2 — registerMcpProxy: let MCP proxy tools declare their structure

> Part of #928 — motivation, the step-by-step identity walkthrough, and
> the rule-author guidance live there. This PR is the registration half.
> **Based on #929** (the derivation fix); review that first — this branch
> is small on top and delegates to the same derivation, so there is exactly
> one implementation.

## What changes

A fact-shaping registration API in the `registerToolAccessExtractor` pattern,
so proxy-shaped MCP clients get `mcp`-surface gating under **any** tool name:

- `src/access-intent/mcp-proxy-registry.ts` — `McpProxyRegistry` (one
  registration per tool name, identity-guarded disposers), the
  `McpInvocation` / `McpProxyRegistration` types, and the built-in descriptor
  for the adapter arg shape
- `createMcpTargetsFromInvocation` in `mcp-targets.ts` — the shared candidate
  tail for the built-in field-read path and registered descriptors
- `normalizeInput` takes an optional `McpProxyLookup`; a registered name
  routes to the `mcp` surface, a declining/throwing descriptor falls back to
  the generic extension surface
- `PermissionManager` options gain `mcpProxyLookup`; `PermissionsService`
  gains `registerMcpProxy` / `getMcpProxy`
- `index.ts` pre-registers the literal `mcp` (pi-mcp-adapter's shape) as the
  built-in default — existing behavior byte-identical

Descriptors return **facts** (verb, tool, server), never rule targets — the
permission system stays the authority on candidates, ordering, and baselines.

Docs: `docs/cross-extension-api.md` gains the interface listing plus three
worked shapes (adapter args under a different name, aggregator with prefixed
names incl. a live `getServers`, custom argument shape).

## Deliberate follow-up (out of scope)

Removing the literal `mcp` special case from `classifyToolKind` entirely,
leaving registration as the only routing — ripples through every exhaustive
`ToolKind` switch and deserves its own change once this API has proven out.

## Tests

New `mcp-proxy-registry.test.ts` (registry lifecycle + built-in descriptor),
four normalizer cases (routing under own name, `getServers` derivation,
decline fallback, unregistered names unchanged). Full package suite:
4261/4261; `tsc --noEmit` and `biome check` clean.


georgeharker:
The Fallow audit added in the recent CI changes flags the test cases added by this PR as `stale` (`input-normalizer.test.ts`, `mcp-targets.test.ts`, `permission-manager-unified.test.ts` — 17 active / 442 stale rows in the audit output).

These exercise the new `mcpProxyLookup` registration path (registered proxy name → mcp surface → server-derived targets → generic-surface fallbacks), so the "stale" classification may be the audit encountering a surface it hasn't catalogued yet rather than genuinely fallow cases.

Two questions:

1. Is there a preferred pattern for introducing new test files/suites so the audit classifies them correctly — or a config scope we should update for the new registration path?
2. Could you clarify what "stale" means in Fallow's model? If it's *test cases whose asserted code-path hasn't been observed since the code last changed*, we may be able to restructure these to conform — but if it's *tests for capabilities the audit hasn't catalogued*, a scope update on the audit side might be the right fix.

Happy to restructure or adjust — the tests themselves are stable (4265/4265 on the branch including these).

gotgenes:
Thanks for this, @georgeharker — and for tidying the stray root-level files, which I never came back to acknowledge.

Closing this one, but the capability is tracked and not dropped: #946 carries gap 2 of #928 with this PR named as its reference implementation.

Two reasons for closing rather than asking you to rebase:

First, the base moved under you. The derivation half landed with a different shape than #929 proposed — 1994b3227e16898c8ca4ff0d7a17b258dc306bcb reversed the candidate order to tool-name-first, and c57369648a4ffd64de9b07999e2687ceaa96e867 routed the whole `mcp` surface through `evaluateAnyValue` and deleted `evaluateFirst`. This PR now conflicts across `mcp-targets.ts`, `permission-manager.ts`, `input-normalizer.ts`, and three test files. That's a substantial rebase of 990 lines.

Second, and the reason I don't want you doing that rebase on spec: `registerMcpProxy` is a public cross-extension registration API, and I want to weigh its shape on its own terms rather than as the tail of a derivation fix — including how it answers the inherited-lookup question `ToolAccessExtractorRegistry` already answers under ADR 0012. That design isn't settled yet, so a rebase now would be work toward a target that may move.

The problem you identified is real and I'm not disputing it: `classifyToolKind` matching the literal name `mcp` means a renamed proxy or an aggregator gateway silently loses per-server rules, tool-pattern rules, and `target` evidence for authorizers. That's in #946 in your framing, with this PR linked.

Your issue write-up in #928 did a lot of work here, well beyond the diff — the per-naming-style candidate tables are what let me verify the defect quickly, and chasing down why an exact-server rule didn't fire is what surfaced the deeper matcher bug underneath it. Both halves shipped in v33.0.0 with your `Co-authored-by:` on the implementing commits.

Refs #928, #946.

Issue #918, opened 2026-09-11
pi-subagents-worktrees: `<project_context>`'s inherited `path="..."` attribute defeats worktree isolation, the same way #640's footer did


aaronkyriesenbach:
## Package

pi-subagents, pi-subagents-worktrees

## Package version

@gotgenes/pi-subagents 21.6.0, @gotgenes/pi-subagents-worktrees 0.3.3 (pi-coding-agent 0.84.2)

## What happened?

A child given an isolated worktree by `pi-subagents-worktrees` still inherits its
parent's `<project_context>` block verbatim, including each context file's
`path="<parent's absolute cwd>/AGENTS.md"` attribute. ADR 0006 places
`<project_context>` inside the Identity region on purpose, for the prefix-caching
reason #180 established — so this isn't a regression of #640's fix, which only
ever targeted the footer (a per-session-resolved layer explicitly *outside*
Identity).

But the practical effect is the one #640 already described: the child's prompt
states two contradictory locations for itself — an early, salient,
machine-structured `path="/Users/me/project/AGENTS.md"` attribute naming the
parent's checkout, and a later, correct `Working directory: <own worktree>` in
its freshly-built per-session block. In a real run (5 subagents, 2 waves, mix of
`engineering-backend-architect`/`engineering-frontend-developer` types, all in
`worktreeAgents`, real multi-step tickets), 4 of 5 children resolved that
conflict by trusting the inherited path and `cd`-ing there for the rest of their
run, walking themselves back into the shared parent checkout exactly as #640
described — confirmed via each child's own transcript
(`~/.pi/agent/sessions/.../tasks/*.jsonl`) and by each session's
actually-assigned worktree directory (recorded in the session's own `cwd`)
sitting unused afterward. The 5th child, on a stronger model, stayed in its
assigned worktree; a weaker/mid-tier model reliably didn't.

No error or indication surfaced in either case — same as #640.

## Expected behavior

A workspace-isolated child shouldn't receive an unqualified absolute-path claim,
anywhere in its prompt, that it could reasonably mistake for its own session's
directory — matching the spirit of #640's fix, without giving up the Identity
region's byte-identical-prefix caching property ADR 0006/0008 are built around
(an in-place rewrite of the cached `path="..."` attribute would cost every
isolated child the shared prefix, the same problem ADR 0008 already fixed once
for the tool-listing case by relocating rather than editing in place).

## Steps to reproduce

[...]

PR #959, opened 2026-09-20
fix(pi-subagents): recognize pi ≥0.86's section-shaped prompt in the tail anchors

georgeharker:
## Problem

pi 0.86 moved `buildSystemPrompt` to tagged sections joined by blank lines, and all three deltas defeated the session-resolved-tail anchors on a relocated child:

1. **The footer is gone.** The cwd renders as a `<cwd>\n<path>\n</cwd>` section, so `Current working directory: ${parentCwd}` never matches — `catalogueCloseBefore` is permanently on its "downstream rewrote Pi's output" best-guess path.
2. **The catalogue is wrapped.** It lives inside a `<skills>` section, so the `<skills>` open tag sits between the context block and the heading. `projectContextStart`'s positional guard requires the block's closing tag to be the last non-blank line above the anchored tail — the wrapper breaks the adjacency and the guard returns −1.
3. **The lead-in moved.** 0.86's `renderProjectContext` drops the blank line below the opening tag, so the lead-in sits at `openAt+1`, not the `openAt+2` the guard checks.

Net effect (peer range is `>=0.81.0`, so 0.86 is in-range): a relocated child's `cutProjectContext` cut (#918) never fires, and the child inherits the parent's `<project_context>` block — naming the parent's files by absolute path — alongside its own re-rendered block.

## Fix

The anchors dispatch on **which cwd layer is present**, never version sniffing:

- The footer path is byte-identical to before (all ≤0.85 fixtures unchanged).
- A `<cwd>` section whose content line is exactly the parent's cwd (and whose closing tag follows) takes the anchor's place — the same whole-line discipline the footer applies.
- From that anchor, the walk-back over the section join finds `</skills>`, validates its opening by Pi's own heading, and cuts at the **opening tag** — dropping the wrapper with the layers it carries and leaving the previous section's closing tag adjacent to the tail for the project-context anchor.
- The project-context guard accepts the lead-in at either renderer's offset (+1 on 0.86, +2 through 0.85), keeping the quoted-opening guard intact.

Byte shapes verified against the 0.86.1 SDK dist (`buildSystemPromptSections`' `<name>\n…\n</name>` wrap and `"\n\n"` join; `renderProjectContext` without the leading blank line; `formatSkillsForPrompt` heading unchanged).

## Verification

- New `pi ≥0.86 section shape` fixtures in `test/session/prompts.test.ts` are hand-built from those dist shapes (the pinned 0.84.4 SDK cannot generate them); all other fixtures still route through Pi's own formatter.
- Killing mutation: reverting the source fails 5 of the 7 new tests; every 0.85 test stays green.
- Full `pi-subagents` suite green locally apart from 4 pre-existing, environment-only failures in `test/config/custom-agents.test.ts` (an ambient `PI_CODING_AGENT_DIR` leaks agent files into `getAgentDir()`; CI does not set it).

## Companion change

The claude-bridge capture mirror (pi-claude-bridge#89) cuts parent prompts at the same markers to derive the keys it matches children under — it moves in lockstep with this one. Today it mirrors the pre-fix behavior (tail cut, dangling wrapper) deliberately; a companion bridge change updates `stripSessionLayers` and the 0.86 fixture once this lands.


gotgenes:
Thanks for this, @georgeharker — and for tracing the byte shapes rather than just describing the symptom. It made the review much faster.

I verified the defect independently before reading the diff for design, against the real 0.86.1 dist rather than a hand fixture:

```bash
pnpm view @earendil-works/pi-coding-agent@0.86.1 dist.tarball
# scratch install, then buildSystemPrompt() with contextFiles + skills + cwd
```

All three deltas reproduce, and feeding that prompt through `main`'s `buildAgentPrompt` with a relocated child gives back the parent's `<project_context>` naming `/parent/AGENTS.md`, plus an unclosed `<skills>` tag on *every* 0.86 child. Your branch handles the real bytes for both the relocated and same-cwd cases, not only its own fixtures. I also confirmed your killing-mutation claim: replaying `main`'s `prompts.ts` under your test file fails exactly 5 of the 7 new tests.

The four `test/config/custom-agents.test.ts` failures you saw did not occur on my machine either, which supports your read that they are an ambient `PI_CODING_AGENT_DIR` leak rather than anything in this change.

Since you had maintainer edits enabled, I pushed two non-behavioral commits to the branch rather than asking you for another round:

- `cae62ab5` renames `tailStart` to `cwdAnchoredTailStart`. It sat directly below `sessionResolvedTailStart` and read as a restatement of it, and it also shadowed the local `tailStart` in `inheritedIdentity`.
- `2c1c1bb0` notes on `PROJECT_CONTEXT_LEAD_IN` that the `openAt + 2` arm is dead once the peer floor moves past 0.85, so it gets dropped with its fixtures rather than inherited.

Both are comment-and-name only; `pnpm run check`, `pnpm run lint`, and the 1803-test `pi-subagents` suite are green on the result. Say the word if you would rather they read differently.

Two adjacent cases of the same 0.86 reshape came out of the review and are filed separately so they do not enlarge this diff:

- #961 — `renderProjectContext` still emits the ≤0.85 block shape, so the block a relocated or `portable` child renders for *itself* no longer matches Pi's.
- #962 — `@gotgenes/pi-permission-system`'s tool-surface relocation anchors on `Current working directory:` / `Available tools:` / `Guidelines:`, none of which 0.86 writes, so a session sees Pi's unfiltered tool list above the narrowed block.

Merging now. Thanks again.

georgeharker:
The renames read better than mine did — `cwdAnchoredTailStart` names the discriminator, and the shadowed local in `inheritedIdentity` was a real catch. Keeping both as landed.

Two notes on the mirror side, now that this is merged:

- The bridge capture (elidickinson/pi-claude-bridge#89, commit `e274a69`) already mirrors exactly these anchors — `<cwd>` section dispatch, the `<skills>` wrapper cut at its opening tag, and the dual lead-in offsets — so lockstep is live as of this landing. It carries both 0.86 cuts as keys (the wrapper cut for post-#959 embeddings, the heading cut for pinned pre-#959 ones) until the release carrying this fix ages the pinned versions out.
- Mirrored your `2c1c1bb0` retirement note onto the bridge's `PROJECT_CONTEXT_LEAD_IN` (pi-claude-bridge#89, commit `6f45d9f`), with the added sentence that the mirror retires its arms together with the embedding side — whoever moves the floor past 0.85 finds both halves.

And on the two follow-ups: #961 and #962 are exactly where I'd have drawn the boundary, and #962's severity framing (over-disclosure direction on the exposure surface) is the right read. Happy to take either one if you want to delegate the bridge-adjacent half of #961 — the fixtures in pi-claude-bridge#89 already encode the 0.86 byte shapes, so the anchor-side verification is largely done.

